PR #7, opened 2021-01-28
Setup file for module installation in local environment

ernie2k:
Inclusion of a generic setup file to identify the Pysages package in a particular virtual environment. This file can be extended in next iterations with testing units (pytest, ...), and additional user-dependent entries.


pabloferz:
Thanks for this. We got an initial version on #13. Any other improvements and PRs are of course always welcome.

PR #13, opened 2021-05-10
Setup.py to install pysages

InnocentBug:


InnocentBug:
Let's talk about it tomorrow. I have some questions.

InnocentBug:
Do you have a reason, why you don't like versioneer?
I don't really know what you intend with a github bot, but versioneer running at install time can do so much more.
It can especially track and notice changes that are not a full new version, which at least helps me significantly when tracking other peoples errors.

pabloferz:
I see what you mean. It's just that versioneer feels too bloated. Would something like https://github.com/jbweston/miniver would work instead?

pabloferz (approved):

pabloferz:
Let's merge and see how it goes. We can always fix any issue later.

InnocentBug:
I address the failing test in #15

PR #15, opened 2021-05-13
Fix tests for setup.py.

InnocentBug:


InnocentBug:
The problem seems to an issue with `plum-dispatch` the installation via pip before hand seems to be a work-around.
Not an ideal one though.

I reported it as an issue to plum-dispatch. https://github.com/wesselb/plum/issues/8
Let's see if there is a response.

InnocentBug:
Oh, and if you merge, please squash the commits.

pabloferz:
Merging for now, but we should certainly revise this.

PR #17, opened 2021-05-14
Split methods in separate files

InnocentBug:
This splits the methods into separate files, with two advantages beyond cleaning.
1) hide inner workings of the methods from the user (`_` prefix) for the functions.
2) easier extension to more methods, just add a separate file for the methods.

Let me know what you think about the direction here.

I haven't tested if it works yet.


InnocentBug:
At the moment, the proposed change breaks existing scripts, because I do not expose the method classes in the `pysages.ssages.methods` namespace.
If backwards compatibility is wanted, we can expose the classes in the namespace though.

pabloferz:
I was thinking on moving what you have in `__init__.py` into a `core.py` file and using `__init__.py` only to explicitly export the default, visible symbols.

We can get away breaking the API for now as the is no official release yet.

InnocentBug:
I agree, I also thought about moving the contents of init into a separate file and importing only in init.
Let me push that.

pabloferz (approved):
LGTM

PR #19, opened 2021-05-18
Add a pytproject.toml file

InnocentBug:
This hopefully defines a good build environment for plum-dispatch.


.github/workflows/quick_tests.yml
```diff
@@ -23,8 +23,26 @@ jobs:
...
+                steps:
+                       - uses: actions/checkout@v2
+                       - uses: actions/setup-python@v2
+                         with:
+                             python-version: '3.x'
+                       - name: install dependecies
+                         run: |
+                           pip install jaxlib
+                       - name: install pysages
+                         run: pip install .
```
pabloferz, line 44: I think it needs to be run with `-e`

InnocentBug:
Ok, so I researched it a bit.
setuptools does not fully support integration with pyproject.toml. See here for their discussion on it. https://github.com/pypa/setuptools/issues/1688

However, what they propose is to eliminate `setup.py` entirely. Which solves our issue, because if you remove the `setup.py` users are forced to install via `pip` directly.

We could specify all the config of `setup.py` in `setup.cfg` and make the build environment in `pyproject.toml`. The only thing that wouldn't work in the current scenario is the automatic version determination via `minivers`.
So if move back to manual specification of the version, I can try it out.

I am not really a fan of this, but maybe it is lesser evil.

pabloferz:
It seems that there are efforts to deprecate installing via `setup.py` anyway. In case we do that and lead only a `pyproject.toml` and `setup.cfg`, we should be able to use this https://github.com/pypa/setuptools_scm/ as alternative to `miniver`, as it looks to be compatible with `pyproject.toml`

.github/workflows/quick_tests.yml
InnocentBug: `-e` is for editable build.
That's a feature for developing, it has to work without it too.

InnocentBug:
Good idea to switch to the setuptools_scm.

I implemented it and could drop the `setup.py` entirely.
A bit ironic, because this is what I originally wanted, but `pyproject.toml` seems to be the future.

pabloferz (approved):

PR #22, opened 2021-05-19
Standardize licenses and copyright notices

pabloferz:


InnocentBug (approved):
There are a few license/contribution unrelated changes included, which isn't ideal, but won't hurt either.

PR #30, opened 2021-06-24
Add method to restore snapshots

pabloferz:
These helper functions (`indices`, `momenta`, `restore`) seem to start screaming for a dedicated data structure.

### DIFF
--- a/pysages/backends/common.py
+++ b/pysages/backends/common.py
@@ -1,0 +1,22 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2020-2021: PySAGES contributors
+# See LICENSE.md and CONTRIBUTORS.md at https://github.com/SSAGESLabs/PySAGES
+
+
+def restore(view, restore_vm, snapshot, prev_snapshot):
+    # Create a mutable view of the jax arrays
+    positions = view(snapshot.positions)
+    forces = view(snapshot.forces)
+    ids = view(snapshot.ids)
+    # Overwrite the data
+    positions[:] = view(prev_snapshot.positions)
+    forces[:] = view(prev_snapshot.forces)
+    ids[:] = view(prev_snapshot.ids)
+    # Special handling for velocities and masses
+    restore_vm(snapshot, prev_snapshot)
+
+
+# Fallback method for restoring velocities and masses
+def restore_vm(view, snapshot, prev_snapshot):
+    vel_mass = view(snapshot.vel_mass)
+    vel_mass[:] = view(prev_snapshot.vel_mass)
--- a/pysages/backends/common.py
+++ b/pysages/backends/common.py
@@ -2,8 +2,22 @@
 # Copyright (c) 2020-2021: PySAGES contributors
 # See LICENSE.md and CONTRIBUTORS.md at https://github.com/SSAGESLabs/PySAGES
 
+from typing import Callable, NamedTuple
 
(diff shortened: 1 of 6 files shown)


pabloferz:
The only information that is being restored right now are the positions, velocities, masses, forces and ids. Not sure what is the minimum of information that we would need to restore in general, though.

pysages/backends/openmm.py
```diff
@@ -169,9 +180,9 @@ def bind(context, sampling_method, force = Force(), **kwargs):
     #
     force.add_to(context)
     wrapped_context = ContextWrapper(context, force)
-    indices, momenta, bias = build_helpers(wrapped_context.view)
+    indices, momenta, bias, restore = build_helpers(wrapped_context.view)
```
InnocentBug, line 183: I see, that we are returning here tuples a couple of times.
Maybe it is more flexible to return here a dictionary or named_tuple instead from the helpers.

If we have to adjust to include more returns, we just add another field to the dict or namedtuple, without having to refactor old code, because it returns more values now.

pysages/backends/common.py
```diff
@@ -0,0 +1,22 @@
...
+    # Create a mutable view of the jax arrays
+    positions = view(snapshot.positions)
+    forces = view(snapshot.forces)
+    ids = view(snapshot.ids)
+    # Overwrite the data
+    positions[:] = view(prev_snapshot.positions)
+    forces[:] = view(prev_snapshot.forces)
+    ids[:] = view(prev_snapshot.ids)
+    # Special handling for velocities and masses
+    restore_vm(snapshot, prev_snapshot)
```
InnocentBug, line 16: I don't understand this section complete.

I understand, that it makes sense to not restore masses and velocities all the time. (some methods/applications don't need it.)
But the restore function, doesn't offer no restore as an option. Am I missing something here?

pysages/ssages/methods/abf.py
```diff
@@ -41,7 +41,7 @@ def _abf(snapshot, cv, grid, N, helpers):
     dims = grid.shape.size
     natoms = np.size(snapshot.positions, 0)
     get_grid_index = build_indexer(grid)
-    indices, momenta = helpers
+    indices, momenta, *_ = helpers
```
InnocentBug, line 44: this is the part, where a namedtuple or dict, wouldn't need refactoring of old code.
(the additional fields would just be ignored.)

InnocentBug (changes requested):
I like the PR. It looks nice and does what we need for FFF.

I have a few comments directly in the code.
They not *must* changes, but might be worth thinking about it.

Can we add an automated test, that checks if this works as expected?
We should add a test for every new functionality we are adding.
(I see some challenges here, but nothing we can't overcome.)

Good work!

pysages/backends/common.py
pabloferz: Well, technically you can pass a `restore_vm` here that effectively doesn't do anything, e.g. `lambda s, ps: None`. The method `restore_vm` defined here is a convenience fallback, but you're right in that there is no way for the user to specify that.

InnocentBug: OK, now I see how the user can easily exchange the method if needed.
And to allow passing a function gives greater flexibility. I like the method.
And passing a NOP lambda should definitely be fine.
When we go and write the documentation, we should probably mention it there.

InnocentBug (approved):
Let's go forward with this now.

PR #32, opened 2021-07-22
More flexibility for jitting `update` methods

pabloferz:
Before this, `generalize` tried to jit the `update` methods with no option to prevent this. Sometimes, we would like some flexibility for jitting only part of the update code, e.g. if we need to raise an exception.


InnocentBug (approved):
looks good to me.
more flexibility is almost never a bad thing :)

PR #34, opened 2021-07-29
Add methods to copy snapshots

pabloferz:


InnocentBug:
Is this ready for review, or do you want to add something.
I think we said the last time that something was missing here.

pabloferz:
Yes, I wanted to add an option to indicate if we want the copy on the CPU if it's on the GPU.

pabloferz:
All right. This is ready for review.

pysages/backends/snapshot.py
```diff
@@ -2,25 +2,49 @@
...
+from pysages.utils import copy
 
 import jax.numpy as np
 
-from collections import namedtuple
 
-
-class Box(namedtuple("Box", ("H", "origin"))):
+class Box(NamedTuple("Box", [
+    ("H",      JaxArray),
```
InnocentBug, line 14: Is "H" really good name for this.
I haven't seen it elsewhere and could we use a better telling name?

tests/test_snapshots.py
```diff
@@ -0,0 +1,39 @@
...
+    old_ptr = old.positions.unsafe_buffer_pointer()
+    new_ptr = new.positions.unsafe_buffer_pointer()
+    old_box_H_ptr = old.box.H.unsafe_buffer_pointer()
+    new_box_H_ptr = new.box.H.unsafe_buffer_pointer()
+
+    # When copying to CPU we get a `numpy.ndarray` instead of a
+    # `jaxlib.xla_extension.DeviceArray`
+    new_cpu = copy(old, ToCPU())
+
+    assert np.all(old.positions == new.positions).item()
```
InnocentBug, line 35: Ok, I was double checking if assert is the right way to test in automated frameworks.
But it seems like that pytest relies on assert, while unittest wants exceptions.

InnocentBug (approved):
Looks good to me.

I am happy that we have a test case now :)

pysages/backends/snapshot.py
pabloferz: Agreed, I'm not sure what to call it, though. It represents the affine transform that needs to be applied to the positions for not orthogonal boxes. Maybe `transform` is good enough?

tests/test_snapshots.py
pabloferz: Do you have a preference on `unittest`. I like that `pytest` doesn't require too much boilerplate.

InnocentBug: no, I don't have a preference. My testing experience is more about C++. I read up on it and seems like it is as you say. both are fine and pytest is easier to work with. So I am happy to go along with it.

pysages/backends/snapshot.py
InnocentBug: `transform` sounds good to me

pabloferz: OK. Will change that in a separate PR. Merging this for now.

PR #37, opened 2021-08-05
Callback

InnocentBug:
Optional callback integrated after update.

And some work to have more inheritance and less duplicate code.
Adding some documentation on the way.

### DIFF
--- a/pysages/backends/core.py
+++ b/pysages/backends/core.py
@@ -22,6 +22,7 @@ def current_backend():
     if _CURRENT_BACKEND is not None:
         return _CURRENT_BACKEND
     warnings.warn("No backend has been set")
+    return None
 
 
 def supported_backends():
@@ -42,15 +43,19 @@ def set_backend(name):
 
 
 def bind(context, sampling_method, **kwargs):
-    """Couples the sampling method to the simulation."""
-    #
+    """Couples the sampling method to the simulation.
+
+    context -- hoomd simulation context
+    sampling_method -- pysages sampling_method
+    callback -- callback method with call signature `callback(snapshot, state)`
+      called after PySages updated. Example: logging of CVs.
+    """
     if type(context).__module__.startswith("hoomd"):
         set_backend("hoomd")
     elif type(context).__module__.startswith("simtk.openmm"):
         set_backend("openmm")
-    #
+
     check_backend_initialization()
-    #
     return _CURRENT_BACKEND.bind(context, sampling_method, **kwargs)
 
 
@@ -60,17 +65,40 @@ def check_backend_initialization():
 
 
 class ContextWrapper:
+    """
+    Abstract base class managing the context of the backends.
+    """
     def __init__(self, context):
+        """
+        Store the context of the simulation to be wrapped.
+        """
         self.context = context
 
+
 class Sampler:
-    def __init__(self, method_bundle, bias):
+    """
+    Abstract base class to facilitate the update calls between backend and pysages.
+    """
+    def __init__(self, method_bundle, bias, callback=None):
+        """
+        Initialize the sampler.
+
+        method_bundle -- tuple with (snapshot, initialize, update) instances to be stored.
+        bias -- bias to be applied to the forces
+        callback -- optional callback function/functor with call signature (snapshot, state) called after updates. Examples: logging user calculated observables or CVs during the run.
+        """
         snapshot, initialize, update = method_bundle
         self.snapshot = snapshot
         self.state = initialize()
         self.update_from = update
         self.bias = bias
-    #
+        self.callback = callback
+
     def update(self):
+        """
+        Execute the update step.
+        """
         self.state = self.update_from(self.snapshot, self.state)
         self.bias(self.snapshot, self.state)
+        if self.callback:
+            self.callback(self.snapshot, self.state)
--- a/pysages/backends/core.py
+++ b/pysages/backends/core.py
@@ -57,3 +57,20 @@ def bind(context, sampling_method, **kwargs):
 def check_backend_initialization():
     if _CURRENT_BACKEND is None:
         raise RuntimeError("No backend has been set")
+
+
+class ContextWrapper:
+    def __init__(self, context):
+        self.context = context
+
+class Sampler:
+    def __init__(self, method_bundle, bias):
+        snapshot, initialize, update = method_bundle
+        self.snapshot = snapshot
+        self.state = initialize()
+        self.update_from = update
+        self.bias = bias
+    #
+    def update(self):
+        self.state = self.update_from(self.snapshot, self.state)
+        self.bias(self.snapshot, self.state)
--- a/pysages/backends/hoomd.py
+++ b/pysages/backends/hoomd.py
@@ -16,28 +16,26 @@ from jax.dlpack import from_dlpack as asarray
 from pysages.backends.common import HelperMethods
 from pysages.backends.snapshot import Box, Snapshot
 
-from .core import ContextWrapper
+from .core import ContextWrapper, Sampler
+
 
 class ContextWrapperHOOMD(ContextWrapper):
+    """
+    ContextWrapper for hoomd-blue.
+    """
     def __init__(self, context):
         super().__init__(context)
         self.sysview = SystemView(context.system_definition)
         self.synchronize = self.sysview.synchronize
 
 
-class Sampler(HalfStepHook):
-    def __init__(self, method_bundle, bias):
-        super().__init__()
-        #
-        snapshot, initialize, update = method_bundle
-        self.snapshot = snapshot
-        self.state = initialize()
-        self._update = update
-        self.bias = bias
-    #
-    def update(self, timestep):
-        self.state = self._update(self.snapshot, self.state)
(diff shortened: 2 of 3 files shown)


pysages/backends/core.py
```diff
@@ -42,18 +43,62 @@ def set_backend(name):
...
 def bind(context, sampling_method, **kwargs):
-    """Couples the sampling method to the simulation."""
-    #
+    """Couples the sampling method to the simulation.
+
+    context -- hoomd simulation context
+    sampling_method -- pysages sampling_method
+    callback -- callback method with call signature `callback(snapshot, state)`
+      called after PySages updated. Example: logging of CVs.
+    """
```
pabloferz, line 52:
```suggestion
    context -- Backend simulation context
    sampling_method -- PySAGES sampling method
    callback -- callback method with call signature `callback(snapshot, state, timestep)`
      called after each pysages update. Example: logging of CVs.
    """
```

Unrelated, but this makes me think that for easy discoverability we should probably add a couple of methods to list available sampling methods and collective variables. We could open an issue to track this.

I think that I would add callback as an explicit argument here:

```python
def bind(context, sampling_method, callback: Callable = None, **kwargs):
   # ...
```

pysages/backends/hoomd.py
```diff
@@ -16,27 +16,26 @@
...
-    def update(self, timestep):
-        self.state = self._update(self.snapshot, self.state)
-        self.bias(self.snapshot, self.state)
+class SamplerHOOMD(Sampler, HalfStepHook):
+    """
+    Sampler class for hoomd-blue backend.
+    """
+    def __init__(self, method_bundle, bias, callback=None):
+        HalfStepHook.__init__()
+        Sampler.__init__(method_bundle, bias, callback)
```
pabloferz, line 38: For HOOMD, `update` needs to be defined and take a `timestep` argument (we are mirroring a C++ class here). I could modify the OpenMM C++ plugin to fit this interface to make them uniform. But for now we need to special case either one or the other.

pysages/backends/core.py
```diff
@@ -42,18 +43,62 @@ def set_backend(name):
...
+        self.callback = callback
+
+    def update(self):
+        """
+        Execute the update step.
+        """
+        self.state = self.update_from(self.snapshot, self.state)
+        self.bias(self.snapshot, self.state)
+        if self.callback:
+            self.callback(self.snapshot, self.state)
```
pabloferz, line 104:
```suggestion
            self.callback(self.snapshot, self.state, timestep)
```

Right now the `update` underlying C++ interfaces for HOOMD and OpenMM differ in that one takes a timestep (HOOMD) and the other doesn't, we should make the callback interface uniform regardless. See also a comment below.

pysages/backends/core.py
```diff
@@ -42,18 +43,62 @@ def set_backend(name):
...
+    """
+    Abstract base class to facilitate the update calls between backend and pysages.
+    """
+    def __init__(self, method_bundle, bias, callback=None):
+        """
+        Initialize the sampler.
+
+        method_bundle -- tuple with (snapshot, initialize, update) instances to be stored.
+        bias -- bias to be applied to the forces
+        callback -- optional callback function/functor with call signature (snapshot, state) called after updates. Examples: logging user calculated observables or CVs during the run.
```
pabloferz, line 88:
```suggestion
        callback -- optional callback function/functor with call signature
          `(snapshot, state)` called after updates. Examples: logging user calculated
          observables or CVs during the run.
```

pysages/backends/hoomd.py
```diff
@@ -125,15 +124,24 @@ def bias(snapshot, state, sync_backend):
...
 
 def bind(context, sampling_method, **kwargs):
-    #
+    """
+    Bind pysages to hoomd-blue.
+
+    context -- hoomd simulation context
+    sampling_method -- pysages sampling_method
+    callback -- callback method with call signature `callback(snapshot, state)`
+      called after PySages updated. Example: logging of CVs.
```
pabloferz, line 133:
```suggestion
    sampling_method -- PySAGES sampling method
    callback -- callback method with call signature `callback(snapshot, state, timestep)`
      called after each pysages update. Example: logging of CVs.
```

pysages/backends/hoomd.py
```diff
@@ -125,15 +124,24 @@ def bias(snapshot, state, sync_backend):
...
+
     wrapped_context = ContextWrapper(context)
+
     snapshot = take_snapshot(wrapped_context)
     method_bundle = sampling_method(snapshot, helpers)
-    sync_and_bias = partial(bias, sync_backend = wrapped_context.synchronize)
-    #
-    sampler = Sampler(method_bundle, sync_and_bias)
+    sync_and_bias = partial(bias, sync_backend=wrapped_context.synchronize)
+    pysages_callback = kwargs.get("callback", None)
```
pabloferz, line 142: If we make callback an explicit argument then this won't be needed.

pysages/backends/openmm.py
```diff
@@ -19,26 +19,27 @@
...
+from .core import ContextWrapper, Sampler
 
-class ContextWrapper:
+
+class ContextWrapperOpenMM(ContextWrapper):
+    """
+    OpenMM derivative of the ContextWrapper class.
+
+    context -- as used in base class
+    force -- OpenMM force class that is wrapped
```
pabloferz, line 30:
```suggestion
    force -- `openmm_dlext` force class that is being wrapped
```

pysages/backends/openmm.py
```diff
@@ -176,17 +177,27 @@ def check_integrator(context):
...
 
-def bind(context, sampling_method, force = Force(), **kwargs):
+def bind(context, sampling_method, force=Force(), **kwargs):
+    """
+    Bind OpenMM as a backend to PySages.
+
+    context -- OpenMM context
+    sampling_method -- pysages sampling method
+    callback -- callback method with call signature `callback(snapshot, state)`
+      called after PySages updated. Example: logging of CVs.
```
pabloferz, line 187:
```suggestion
    Bind OpenMM as a backend to PySAGES.

    context -- OpenMM context
    sampling_method -- PySAGES sampling method
    callback -- callback method with call signature `callback(snapshot, state, timestep)`
      called after each pysages update. Example: logging of CVs.
```

pysages/backends/openmm.py
```diff
@@ -176,17 +177,27 @@ def check_integrator(context):
...
+
     wrapped_context = ContextWrapper(context, force)
     helpers, bias = build_helpers(wrapped_context.view)
     snapshot = take_snapshot(wrapped_context)
     method_bundle = sampling_method(snapshot, helpers)
-    sync_and_bias = partial(bias, sync_backend = wrapped_context.synchronize)
-    #
-    sampler = Sampler(method_bundle, sync_and_bias)
+    sync_and_bias = partial(bias, sync_backend=wrapped_context.synchronize)
+    pysages_callback = kwargs.get("callback", None)
```
pabloferz, line 198: As above, this won't be needed if we make `callback` an explicit argument to all `bind` methods.

pabloferz (changes requested):

pysages/backends/hoomd.py
InnocentBug: Yes, that is an issue, that hoomd and openmm differ here.
But I still don't think the callback needs an explicit timestep.
Doesn't snapshot have an attribute for timestep?
And if it doesn't, we should add one.

pysages/backends/core.py
InnocentBug: See comment above. I don't think we need an explicit timestep for the callback.

pysages/backends/hoomd.py
InnocentBug: sure, we can make it an explicit argument.
I didn't do it, because it more typing, but if improves the design, I'll change it.

InnocentBug:
So, I changed the signature and made the callback with a type.
At the moment, I do not specify the call signature. We can do that later when we have a base class for the state.

pysages/backends/openmm.py
InnocentBug: fixed

pysages/backends/hoomd.py
InnocentBug: fixed

pysages/backends/core.py
InnocentBug: So, I decided to pass the timestep now to the callback.
Maybe it helps a user, idk.

But that requires now, that OpenMM passes the timestep too for the update method

pysages/backends/hoomd.py
InnocentBug: I am strongly voting to unify this somehow.

InnocentBug:
Does it work if I just give `timestep=0` as default argument, so hoomd can set the timestep and openmm ignores it.
Just for now?

pabloferz:
Yes, setting `timestep = 0` is a good compromise.

pysages/backends/hoomd.py
pabloferz: I will look into it, but that has to be done on the `openmm-dlext` side. I opened https://github.com/SSAGESLabs/openmm-dlext/issues/6 to remind us.

PR #39, opened 2021-08-10
Update for jax 0.2.15

pabloferz:
The `from_dlpack` method now detects automatically the device.


InnocentBug (approved):
looks good and works :+1:

PR #41, opened 2021-08-17
More OpenMM on CPU fixes

pabloferz:
When on CPU make sure the inverse masses have the same shape as when on the GPU.


InnocentBug (approved):

pabloferz:
I will add tests after the new testing setup from #34 gets in.

PR #42, opened 2021-09-02
New nn API

pabloferz:
This is a full rewrite of the `nn` submodule. The idea is to merge this in, then switch all the methods that use neural nets to use this instead and then to remove the previous submodule.

### DIFF
--- a/pysages/methods/core.py
+++ b/pysages/methods/core.py
@@ -11,11 +11,13 @@ from jax import jit
 
 from pysages.backends import ContextWrapper
 from pysages.collective_variables.core import build
+from pysages.utils import identity
 
 
 # ================ #
 #   Base Classes   #
 # ================ #
+
 class SamplingMethod(ABC):
     snapshot_flags = set()
 
@@ -99,7 +101,7 @@ def generalize(concrete_update, helpers, jit_compile = True):
     if jit_compile:
         _jit = jit
     else:
-        def _jit(x): return x
+        _jit = identity
 
     _update = _jit(concrete_update)
 
--- a/pysages/ml/__init__.py
+++ b/pysages/ml/__init__.py
@@ -1,0 +1,9 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2020-2021: PySAGES contributors
+# See LICENSE.md and CONTRIBUTORS.md at https://github.com/SSAGESLabs/PySAGES
+
+from . import models
+from . import objectives
+from . import optimizers
+from . import training
+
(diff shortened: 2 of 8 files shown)


pysages/_nn/objectives.py
```diff
@@ -0,0 +1,283 @@
...
+    prod,
+    sum_squares,
+    unpack,
+)
+
+import jax.numpy as np
+import numpy as onp
+
+
+# Losses
```
InnocentBug, line 22: Are these all abstract classes, or why are they all defined here, but not implemented.
I am a little confused here.

pysages/_nn/objectives.py
```diff
@@ -0,0 +1,283 @@
...
+class L2Regularization(Regularizer, NamedTuple):
+    coeff: Float
+
+
+class VarRegularization(Regularizer):
+    pass
+
+
+@dispatch
+def build_cost_function(model, loss: Loss, reg: Regularizer):
```
InnocentBug, line 61: this function overloading is a bit confusing to me.
Is it really the best way to have overloaded `build_cost_function` with different number of arguments.
Wouldn't it make sense to have a different name for the ones that take a model?

pysages/_nn/objectives.py
```diff
@@ -0,0 +1,283 @@
...
+    cost = build_cost_function(loss, reg)
```
InnocentBug, line 62: I don't understand, why this is generated here.
Does it have a side effects?
I don't see where `cost` would be used after.

pysages/_nn/objectives.py
```diff
@@ -0,0 +1,283 @@
...
+        prediction = model.apply(params, inputs).reshape(reference.shape)
+        e = np.asarray(prediction - reference, dtype = np.float32).flatten()
+        ps, _ = unpack(params)
+        return cost(e, ps)
+
+    return objective
+
+
+@dispatch
+def build_cost_function(model, loss: Sobolev1Loss, reg: Regularizer):
```
InnocentBug, line 74: I guess, for all of them. It wouldn't hurt to have a doc string, that explains what the model is and how it is implemented here.

InnocentBug (changes requested):
Some of the implementations confuse me.
But maybe that is my limited understanding of how dispatch is supposed to be used.
Or some documentation could help.

I also think, that an example that uses the code would make it easier for users to get into the code.

InnocentBug (approved):
I can't really test it. But it looks good.

pabloferz:
I'll add a README and tests in a follow up PR.

PR #45, opened 2021-09-29
[HOOMD] Add a global pool of Sampler objects

pabloferz:
We will keep all created samplers around unless they are detached from the corresponding contexts.

### DIFF
--- a/pysages/backends/hoomd.py
+++ b/pysages/backends/hoomd.py
@@ -15,6 +15,12 @@ from hoomd.dlext import (
 from jax.dlpack import from_dlpack as asarray
 from pysages.backends.common import HelperMethods
 from pysages.backends.snapshot import Box, Snapshot
+from warnings import warn
+
+
+# TODO: Figure out a way to automatically tie the lifetime of Sampler
+# objects to the contexts they bind to
+SAMPLERS_CONTEXTS = {}
 
 
 class ContextWrapper:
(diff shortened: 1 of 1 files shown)


pysages/backends/hoomd.py
```diff
@@ -129,4 +135,25 @@ def bind(context, sampling_method, **kwargs):
     sampler = Sampler(method_bundle, sync_and_bias)
     context.integrator.cpp_integrator.setHalfStepHook(sampler)
     #
+    SAMPLERS_CONTEXTS.append((sampler, context))
```
InnocentBug, line 138: if we make this a dictionary, with the context as key and sampler as value, much of the logic for detach would be simpler, right?

InnocentBug: we just have to have a logic to make sure that every context has only one sampler, or we need a list of samplers as values

pabloferz: Good point. This can be handled in bind I think. As HOOMD only accepts one `HalfStepHook` at a time.

pabloferz: Wait. This is fine, `setHalfStepHook` replaces previously bound samplers. And since we are using a dict now, reassigning the entry will also remove the reference to any previous sampler.

InnocentBug (dismissed):

InnocentBug (approved):
Looks good to me and I can confirm, that I don't have to assign the result of `bind` in my script for it to work.

PR #48, opened 2021-10-04
Split the docker conainer in multiple instances

InnocentBug:
This enables easy pull of prebuild containers and we can test inside the container.

This is more a test version so far. I uses my personal docker hub account. An auto-build solution would be better.
Maybe even splitting the plugin builds into the plugin git repos. But that costs money.
So for now, I am experiencing with this.


pysages/ssages/methods/umbrella.py
```diff
@@ -0,0 +1,49 @@
...
+        self.kspring = np.asarray(kspring)
+        self.center = np.asarray(center)
+
+    def __call__(self, snapshot, helpers):
+        return _umbrella(snapshot, self.cv, self.center, self.kspring, helpers)
+
+
+def _umbrella(snapshot, cv, center, kspring, helpers):
+    natoms = np.size(snapshot.positions, 0)
+    indices = helpers.indices
```
pabloferz, line 37:
```suggestion
    def __call__(self, snapshot, helpers):
        return _umbrella(self, snapshot, helpers)


def _umbrella(method, snapshot, helpers):
    cv = method.cv
    center = method.center.reshape(1, -1)
    kspring = self.kspring.reshape(1, -1)
    natoms = np.size(snapshot.positions, 0)
    indices = helpers.indices
```

Does this (along with the previous matmul change) fix the issue in higher dims than one?

PR #61, opened 2021-10-06
Second attempt to implement a callback machamism

InnocentBug:
Less ambitious, but it works.

Note that the testing examples generates automatically histograms as needed for umbrella integration.


InnocentBug:
Yay :partying_face: the CI with the docker container caught an error, that we might have overlooked otherwise.

Issue #38, opened 2021-08-06
Wrapping coordinates


InnocentBug:
Decide if coordinates should be delivered in the snapshots wrapped or unwrapped.
Or make it a user option.

At the moment, results between different backends might differ.

PR #63, opened 2021-10-07
Add CI with docker container

InnocentBug:
The test itself get an improvement with #61 .
Unfortunately the test fails sometimes now.
The reason is #38


InnocentBug:
can we merge this, than I would start the runner implementation based on this

PR #72, opened 2021-10-15
Update OpenMM to version 7.6

InnocentBug:
After merge, we have to manually trigger the build to update docker-hub.
(And for the plugins too.)


pabloferz:
This is going to fail on the `alanine-dipeptide_openmm.py` example, but we should merge it any way. It should hopefully work again after https://github.com/SSAGESLabs/openmm-dlext/pull/11 gets merged.

PR #75, opened 2021-10-19
Wrap coordinates as a snapshot is taken.

InnocentBug:
* hoomd-blue only.
* help for openmm wanted.

### DIFF
--- a/pysages/backends/common.py
+++ b/pysages/backends/common.py
@@ -6,7 +6,6 @@ from typing import Callable, NamedTuple
 
 
 class HelperMethods(NamedTuple):
-    indices: Callable
     momenta: Callable
     restore: Callable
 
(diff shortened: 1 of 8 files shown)


InnocentBug:
Thank you @mquevill for sharing your experience with us.
We talk about it here, because the example is using absolute position in the box as CV, and using this wrapped is not smooth and leads to issues.
I absolutely agree, that distances in groups should be handles with minimum Image condition, so the CV calculation should ensure this.
But we cannot unwrapped positions for the CVs that need absolute coordinates if the backend does not provide the necessary information.

Another option would be to just expose the images in a snapshot as well, than the CV could take care of it.
How does OpenMM handle this, can we expose the images?

pabloferz:
Sure, let's discuss it. I'm all for cleaning all APIs external and internal, if it helps user/developers to understand what's going on.

PR #81, opened 2021-11-04
New hoomd backend

InnocentBug:
Works with the new hoomd-dlext implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,8 @@ GPU support.
 
 ## Usage
 
+Let's update
+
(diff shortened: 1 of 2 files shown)


InnocentBug:
it works :partying_face:

InnocentBug:
we can and should merge this now

pabloferz:
Are the containers ready?

InnocentBug:
yes, they are

I initiated it yesterday and the test would fail otherwise

pabloferz (approved):

PR #85, opened 2021-12-07
Setup sphinx to generate a documentation

InnocentBug:
At the end, we want the API documented.

My stubs docsumentation to keep PR's short.
And later add full API documentation, installation, examples, jupyter books and auto build for read the docs.


InnocentBug:
replaced it with #89

PR #86, opened 2021-12-08
Add docstrings for coordinate CVs

ricalessandri:
I'm going through the code and I started adding missing docstrings. Starting with something simple :)


ricalessandri:
And do we want the docstrings to be simple or do we want to adhere to (more verbose) standards such as "Numpy Style", etc. (the ones they talk about e.g. here: https://www.datacamp.com/community/tutorials/docstrings-python)?

pabloferz:
Thanks @ricalessandri! I think we want verbose docstrings, even if the most important parts of the documentation in the end are going to be tutorials and examples.

pabloferz:
@ricalessandri take a look at the documentation https://pysages.readthedocs.io/en/latest/module-pysages-methods-umbrella.html for an example of the format we are aiming for.

pabloferz (approved):
Let's merge this and we can improve upon it later.

InnocentBug:
I am just learning about sphinx.
We should follow their exact description of how docstrings have to look like.

https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html

InnocentBug:
ok, we are aiming for this instead

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html#example-numpy

ricalessandri:
> ok, we are aiming for this instead
> 
> https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html#example-numpy

OK, I updated those docstrings following this example (at least I think). Take a look.

pysages/collective_variables/coordinates.py
```diff
@@ -8,10 +8,40 @@
...
+    Returns the geometric center, or centroid, of a group of points in space.
+
+    Parameters
+    ----------
+    positions : ndarray
+        Array containing the positions of the points for which to compute the barycenter.
+
+    Returns
+    -------
+    barycenter : ndarray
```
pabloferz, line 21: In all of these `np` is actually `jax.numpy` so the return type is `DeviceArray` (or `jaxlib.xla_extension.DeviceArray`, but that is too verbose)

pabloferz: Strictly speaking, for our collective variables all the inputs and outputs are actually `DeviceArray`s.

InnocentBug (approved):
As a first step, it looks good to me.

pabloferz:
Thank you @ricalessandri

PR #87, opened 2021-12-09
Try handling current deprecations from jax and openmm

pabloferz:
These can be annoying for the user.


InnocentBug (dismissed):
Ok, but there is a lot more going on in this PR.
Do you also have a beautification applied here?

pabloferz:
Yes. I applied black and took the chance to clean the examples a bit.

InnocentBug (approved):
LGTM

PR #88, opened 2021-12-09
Fix copying of snapshots

pabloferz:
Make sure we are getting copies of `DeviceArray`s


pabloferz:
It seems that the previous implementation does not always give a copy across different `jax` versions.

InnocentBug (approved):
LGTM

PR #90, opened 2021-12-10
Make ml compatible with jax's omistaging

pabloferz:
In more recent versions of jax the `pack` method in `ml/utils.py` will fail to jit compile and needs to be made compatible with jax's [omnistaging](https://github.com/google/jax/blob/main/design_notes/omnistaging.md).

This fixes such issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,11 +26,23 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
+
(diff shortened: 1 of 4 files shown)


InnocentBug (dismissed):
I assume, the old one failed?
Can we have a test that would catch that?

pabloferz:
It fails only with more recent jax's versions. But, yes, we should be able to test this. I will add a test.

InnocentBug (approved):
thanks for adding the test

PR #92, opened 2022-01-11
Improve the PySAGES/OpenMM ABF alanine dipeptide example

ricalessandri:
This is only a 1st commit to let you know I started to work on enriching this example.
Starting from adding some output writing.

### DIFF
--- a/examples/abf/alanine-dipeptide_openmm.py
+++ b/examples/abf/alanine-dipeptide_openmm.py
@@ -54,6 +54,9 @@ def generate_simulation(
     simulation = app.Simulation(topology, system, integrator)
     simulation.context.setPositions(positions)
     simulation.minimizeEnergy()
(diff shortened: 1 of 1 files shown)


examples/abf/alanine-dipeptide_openmm.py
```diff
@@ -54,15 +54,18 @@ def generate_simulation(
     simulation = app.Simulation(topology, system, integrator)
     simulation.context.setPositions(positions)
     simulation.minimizeEnergy()
+    simulation.reporters.append(app.PDBReporter('output.pdb', 1000)) # save traj every 1000*dt = 2 ps
+    simulation.reporters.append(app.DCDReporter('output.dcd', 1000)) # save traj every 1000*dt = 2 ps
+    simulation.step(5000) # run for 5000*dt = 10 ps
```
pabloferz, line 59: Is this meant to be run before the sampling starts? The way we have set up things is that pysages handles the number of time steps within the call to `method.run`

PR #94, opened 2022-01-11
Improve FUNN implementation

pabloferz:
This uses the ml instead of the nn submodule for the machine learning part of FUNN. The PR also adds some options to make the method more configurable, for instance, it allows the user to chose the training frequency, which we were missing.

TODO:

 - [ ] Add docs


InnocentBug (approved):
I pushed a commit for some typos and the position of doc strings for sphinx.

Now it looks great!
Good job.

Issue #97, opened 2022-01-18
New Example: OpenMM+Umbrella Sampling


meschw04:
Hello!
I am trying to run OpenMM with umbrella sampling on just a very simple system, then move up to classical alanine dipeptide methods (both via umbrella sampling). I am fairly used to using OpenMM, although I am learning hoomd as well. Below is the system I am trying to run with (originally from Laio 2004):
```
import pysages
from pysages.collective_variables import Component
from pysages.methods import UmbrellaIntegration, ABF
import numpy as np
from openmm import *
from openmm.app import *
from openmm.unit import *

GLOB_TEMP = 0.12028 #Set such that $\beta = 1$
system = System()
system.addParticle(1)
force = CustomExternalForce('-5*e^(-(x/1.75)^2)')
force.addGlobalParameter("e", math.e)
integrator = LangevinIntegrator(GLOB_TEMP,0.0005,0.1)

force.addParticle(0,[])
system.addForce(force)
topology = Topology()
chain = topology.addChain()
residue = topology.addResidue('BAH',chain)
topology.addAtom('C',element.carbon,residue)
simulation = Simulation(topology,system,integrator)

cv1 = CustomExternalForce('x')
cv1.addParticle(0,[])
x = BiasVariable(cv1,-4,4,biasWidth=0.2,periodic=True,gridWidth=100)

cvs = np.array([Component([0], 0)])
method = UmbrellaIntegration(cvs)

centers = list(np.linspace(-4,4,10))
result = method.run(simulation, int(1e5), centers, \
                    50., int(50), int(1e-4))
```
Right now, if I try running this I get:
```TypeError: 'Simulation' object is not callable```
I am by no means an expert in OpenMM, but upon reviewing the ABF+OpenMM example, I thought that this should work with calling the `Simulation` object. Should I not be introducing a `BiasVariable` via OpenMM for the potential?
Thanks so much,
Marcus Schwarting

InnocentBug:
Hey Marcus, thanks for trying out OpenMM and with Pysages.

We have to take down the code example from the readme. It is not upto date anymore.
Instead, try to work with the examples. For OpenMM we have an up-to-date example for the ABF method.

The general scheme is changed in a way, that we need to generate the simulation context with a user function.
And this user function is the first argument that `method.run()` takes.
So just wrap your lines 9-21 in a function and let this function return the simulation object.
And this function is the first argument than.

In the mean time, I will submit a PR that changes the README.

[...]

PR #98, opened 2022-01-18
Update README with ABF example.

InnocentBug:
Maybe we should also have a hoomd example here?


README.md
```diff
@@ -20,45 +20,87 @@ dynamics engine
 
 PySAGES also depends on [JAX](https://github.com/google/jax/), follow their installation
 guide to set it up.
+It is important to install your flavor of jaxlib before you install PySAGES.
+Make sure that you install the CUDA compatible flavor for your system or the CPU version, depending on your local setup.
+PySAGES will not install jaxlib.
```
pabloferz, line 25:
```suggestion
PySAGES also depends on [JAX](https://github.com/google/jax/), follow their installation
guide to set it up. *NOTE:* make sure you manually install jaxlib before PySAGES.
Depending on your local setup, you will have to install the jaxlib CPU version or the CUDA compatible flavor.
```

pabloferz:
In the long run we should remove the examples from the README and link to the documentation instead. For now we could point the user to the examples folder for more examples (including HOOMD).

InnocentBug:
I am not opposed to that idea.
I just think, having a short and nice example on the README can attract some users.
(I always like that.)
Ideally, embedding the example files would be perfect, but it seems like Github does not allow that.

pabloferz:
I'd like to avoid cramping the README with too much information. But we should write a couple of jupyter notebooks and add (Google Collab for example) badges for opening them from the README.

InnocentBug:
Sure, sounds good.
So for now, I would either say, we remove the example from the README?

pabloferz:
We can remove the example and have the following:
```diff
- Here is an example of how to hook pysages to an OpenMM simulation of alanine dipeptide
+ If you want to learn more, you can also head over to the `examples` subfolder,
+ which contains scripts for running different HOOMD-blue and OpenMM biased simulations.
```

InnocentBug:
Sure, maybe we explicitly point to one OpenMM and one HOOMD example.

In #97 Marcus found the HOOMD-blue example for umbrella sampling, but missed the OpenMM ABF example and used the README instead.
So we can point people exactly where they want to start. (Not just method but also Backend.)

Or we sort the examples by backend first and then by method.

pabloferz:
We should definitely sort them. I'd say Backend -> Method -> System is a good way to organize them.

pabloferz:
Let's fix this by removing the example and point generically to `examples` and when we organize the examples we amend this to make it more specific.

InnocentBug:
Sounds good to me.
Both comments

pabloferz (approved):

PR #101, opened 2022-01-25
Forward Flux Sampling

gustavor101:
The ffs is working for openmm


pysages/methods/ffs.py
```diff
@@ -0,0 +1,314 @@
...
+
+    for i in range(0, Num_window0):
+        print(f"Initial stored configuration: {i}\n")
+        helpers.restore(sampler.snapshot, initial_snapshots[i])
+        xi, _= cv(helpers.query(sampler.snapshot))
+        print(xi)
+
+        has_reached_A = False
+        while not has_reached_A:
+            # this can be used not every timestep
```
pabloferz, line 257: It's not clear what is meant by this. Does it mean that we could change it so the argument to `run` is a different number of time steps?

gustavor101: Yes, we can include an argument in run similar to sampling_steps_basin in order to monitor the CV not every timestep in this function.

PR #102, opened 2022-01-26
Google colab integration

InnocentBug:
Integrate the examples on Google collab.

The links of the badges don't work yet, since they are set as if the branch has been merged already.


InnocentBug:
I uploaded all the notebooks.
Do you want to merge this first and then have people test it, or should we ask for test before we merge this?

pabloferz:
I'd like to have others to test them first if possible so we can fix any issues in case people have trouble with them.

InnocentBug:
Checkout this diff on the commit.

https://github.com/SSAGESLabs/PySAGES/pull/102/commits/13cffc4287cbe9bbe7fe797221a4f65c578ac5d6

It looks reasonably reviewable to me.

InnocentBug:
Thanks @ricalessandri for checking out and the Colabs.
Your feedback is very much appreciated.

For comments, I think the best way going forward this, that you comment directly (maybe even with suggestion) here on the PR what you think could be better.
Github let's you directly click on the lines to add the comment where it should go.
For the colabs, it is probably best if you comment on the .md version. Especially if it is text.
They are better for humans to read and to diff.
And I can sync them up later offline.

Also, if it is minor changes, that don't need a discussion, just push and update.

InnocentBug:
Please make sure to add your comments for this one before next Tuesday Feb. 15th.
That we can integrate these soonish.
Thank you.

docs/notebooks/Harmonic_Bias_PySAGES_OpenMM.md
```diff
@@ -0,0 +1,206 @@
...
+
+    simulation = app.Simulation(topology, system, integrator)
+    simulation.context.setPositions(positions)
+    simulation.minimizeEnergy()
+
+    return simulation
+```
+
+<!-- #region id="YtUoUMEdKtH8" -->
+The next step is to define the collective variable. In this case, we choose the two dihedral angles on the molecule as defined by its atom positions. We also choose the fixed center point for the harmonic bias simulation and the corresponding spring constant.
```
ricalessandri, line 158: "fixed center point" is a slightly weird formulation for me. What about "We also choose the equilibrium value at which to constrain the dihedrals...."

docs/notebooks/Umbrella_Integration_PySAGES_HOOMD.md
```diff
@@ -0,0 +1,281 @@
...
+```bash id="rEsRX7GZNJ_R"
+
+rm -rf PySAGES
+git clone https://github.com/SSAGESLabs/PySAGES.git &> /dev/null
+cd PySAGES
+pip install -q . &> /dev/null
+```
+
+<!-- #region id="IXryBllMNiKM" -->
+We can see, that the particle positions are indeed centered around the constraints we set up earlier. Also, we see the shape of the histograms is very similar to the expected analytical prediction. We expect this since a liquid of soft particles is not that much different from an ideal gas.
```
ricalessandri, line 90: leftover text that should not be here

docs/notebooks/Umbrella_Integration_PySAGES_HOOMD.md
```diff
@@ -0,0 +1,281 @@
...
+<!-- #endregion -->
+
+<!-- #region id="LjPjqVjSOzTL" -->
+# Umbrella integration
+
+In the previous section, we demonstrated how PySAGES can be used to run a single simulation with a biasing potential.
```
ricalessandri, line 96: "In the previous section" refers to the other tutorial. Maybe Change to "In the previous tutorial" and add link (or at least the name) to that tutorial?

docs/notebooks/Umbrella_Integration_PySAGES_HOOMD.md
```diff
@@ -0,0 +1,281 @@
...
+
+import pysages
+from pysages.collective_variables import Component
+from pysages.methods import UmbrellaIntegration
+```
+
+```python id="RsZhjfm2U5ps"
+param1 = {"A": 0.5, "w": 0.2, "p": 2}
+
+def generate_context(**kwargs):
```
ricalessandri, line 177: We could think of including a one-liner docstring in functions such as this. This applies also to other notebooks.

ricalessandri (commented):
I've added minor edits by opening a PR on @InnocentBug 's fork. That seemed the right/only way to do :) See https://github.com/InnocentBug/PySAGES/pull/1. I guess if you are fine with those and merge they should show up in this PR.

I further made some relatively less minor suggestions here.

InnocentBug:
I updated the structure and put the notebooks in the example subfolder.
Added a Readme there and organized them by MD engine.
The global readme now links to the example readme

pabloferz (approved):

Issue #54, opened 2021-10-05
[Parallelization] Design API


pabloferz:
Not clear how the API would look like. Any suggestions here are welcome.

pabloferz:
@InnocentBug, @johnaparker

johnaparker:
For parallel execution to work, the MD execution has to moved to a global function, where the inputs and outputs are pickle-able objects. One way to accomplish this is to pass `sampler_args` and `context_args` to generate the Sampler and Context objects, e.g., for Umbrella Integration:
```python
# Global definition
def run_umbrella_sampling(sampler_args, context_generator, context_args, center, kspring, hist_period, hist_offset, timesteps):
    sampler = UmbrellaIntegration(**sampler_args)
    sampler.center = center
    sampler.kspring = kspring
    callback = HistogramLogger(hist_period, hist_offset)

    context = context_generator(**context_args)
    wrapped_context = ContextWrapper(context, sampler, callback)

    with wrapped_context:
        wrapped_context.run(timesteps)

    mean = callback.get_means()
    nabla_A = free_energy_gradient(sampler.kspring, mean, center)
    return dict(kspring=kspring, center=center, histogram=callback, histogram_means=mean, nabla_A=nabla_A)
```

This function can then be dispatched to some parallel scheme, e.g., Dask or MPI Pool:
```python
futures = []
with MPIPoolExecutor() as executor:
    for rep in range(Nreplica):
        context_args["replica_num"] = rep
        sampler_args = dict(cvs=self.cvs)
        futures.append(executor.submit(run_umbrella_sampling, sampler_args, context_generator, context_args, centers[rep], ksprings[rep], hist_periods[rep], hist_offsets[rep], timesteps[rep]))
    for future in futures:
        for key,val in future.result().items():
            result[key].append(val)
```
This approach is implemented in 0d8be39

We'll need to think about if this is the cleanest way to do this. It's a little strange that the sampler object's `run` function is itself spawning sampler objects.

InnocentBug:
The original idea that we had with the context generation was exactly to avoid this situation.
Just have a function that generates every non-picklable object.

Maybe as an idea we abstract the sampling objects and CVS a bit more.
Such that we have a user facing class, that behaves as before, but doesn't acutally initialize anything with the backend.
(So basically just a shell, that stores the arguments the user provided.)
And a second one, that can be generate the actual instances at the point of execution after deployment, by fetching the arguments fro the first one.

[...]

Issue #99, opened 2022-01-19
Completed OpenMM Simulation FES


meschw04:
Hello!
Another question for you all. I noticed from the umbrella integration example with HOOMD, I can take `results = method.run(...)` and use this for getting at the final free energy surface (FES):
https://github.com/SSAGESLabs/PySAGES/blob/192e7f9af6fdb50329d2e8fea095537b39a1fc12/examples/umbrella_integration/integration.py#L111
I got the alanine dipeptide ABF working from the example, however in this case I just run the simulation:
https://github.com/SSAGESLabs/PySAGES/blob/192e7f9af6fdb50329d2e8fea095537b39a1fc12/examples/abf/alanine-dipeptide_openmm.py#L78
however if I set `results = method.run(...)`, then I find that the new variable `results` is a NoneType. Is there a reason that I can set `results=` for a HOOMD simulation, and not `results=` for an OpenMM simulation to get at the FES? How should I get to the FES using the OpenMM alanine dipeptide example?

Thanks so much!

InnocentBug:
The reason you are seeing this, is that we ABF does not return anything after running the simulation.
I am guessing that you have to log something during your run to get the free energy result out of the ABF.
But @pabloferz can help with this probably better than I can.
Maybe you have to write a Callback for ABF to get the results?

However, you should be able to use UmbrellaIntegration also with OpenMM.
You part about writing the CVS and method should be identical between OpenMM and HOOMD-blue.
That is what PySAGES is all about: advanced sampling abstract from the actual MD engine.
So you can try to use UmbrellaIntegration with OpenMM as well.
The reported results should be the same.

pabloferz:
The current way to do this is by manually building a Logger to collect the information and then integrate the forces to obtain the FES (we have some tools to do this). I can try to build an example for you to get the FES @meschw04.

Irrespectively of this, I think we should add the context as attribute to the sampling methods, and a mechanism to get copy of the snapshot as we had already discussed (actually, I'm working on this)

InnocentBug:
I believe this is done now with #135

PR #103, opened 2022-01-26
Make the context part of the method

pabloferz:
This is a stopgap solution to enable building FES for methods such as ABF, but I'm exposing the all `ContextWrapper`s generated in all of our current methods.

This is most likely to change in light of #54, but lets move on with this for now.

### DIFF
--- a/pysages/methods/core.py
+++ b/pysages/methods/core.py
@@ -67,9 +67,9 @@ class SamplingMethod(ABC):
             `kwargs` gets passed to the backend `run` function.
         """
         context = context_generator(**context_args)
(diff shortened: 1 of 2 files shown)


pabloferz:
Should also help fixing #99.

InnocentBug (approved):
It works for now, but for #54 we have to come up with something better.
Probably something pickle-able.

pabloferz:
> It works for now, but for #54 we have to come up with something better. Probably something pickle-able.

Yes, I'm aware. Already thinking on alternatives.

PR #104, opened 2022-02-09
Parallelization API

### LINKED ISSUE #54: [Parallelization] Design API

pabloferz:
Fixes #54.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ repos:
   hooks:
   - id: black
 - repo: https://github.com/mwouts/jupytext
-  rev: v1.13.7
+  rev: v1.13.8
   hooks:
   - id: jupytext
     args: [--sync]
--- a/examples/Install_PySAGES_Environment.md
+++ b/examples/Install_PySAGES_Environment.md
@@ -7,7 +7,7 @@ jupyter:
       extension: .md
       format_name: markdown
       format_version: '1.3'
-      jupytext_version: 1.13.7
+      jupytext_version: 1.13.8
   kernelspec:
     display_name: Python 3
     name: python3
--- a/examples/hoomd-blue/Butane_ANN.md
+++ b/examples/hoomd-blue/Butane_ANN.md
@@ -7,7 +7,7 @@ jupyter:
       extension: .md
       format_name: markdown
       format_version: '1.3'
-      jupytext_version: 1.13.7
+      jupytext_version: 1.13.8
   kernelspec:
     display_name: Python 3
     name: python3
--- a/examples/hoomd-blue/Butane_FUNN.md
+++ b/examples/hoomd-blue/Butane_FUNN.md
@@ -7,7 +7,7 @@ jupyter:
       extension: .md
       format_name: markdown
       format_version: '1.3'
-      jupytext_version: 1.13.7
+      jupytext_version: 1.13.8
   kernelspec:
     display_name: Python 3
     name: python3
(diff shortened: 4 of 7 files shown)


pabloferz:
@ndtrung81 I see no issue in adding that. That needs to be done in a separate PR as this is intended for supporting all methods and not Umbrella Sampling in particular. Would you mind opening a separate issue for that specific feature?

pabloferz:
This introduces some changes that should be separate PRs, I'm thinking on haveing:

 - [ ] Make `run` a  module method instead of a class one.
 - [ ] Provide an `analyze` method for post-run processing
 - [ ] Ensure all methods are picklable
 
after having these in `main` we can rebase and clean and this up.

johnaparker:
@pabloferz testing with SerialExecutor seems to be working. Using MPIPoolExecutor, there are two issues right now:

1. Difference between `with executor as ex` and `with executor() as ex`. It might be better to prefer the latter, so the user passes in `SerialExecutor` or `MPIPoolExecutor` without the parentheses.
2. The MPI child processes are not running, and I believe the issue is that `method` cannot be pickled. Inside `umbrella_integration.py`, I'm trying the following:
    ```python
    with open('out.pkl', 'wb') as f:
        pickle.dump(method, f)
        # OR pickle.dump(method.subsamplers[0], f)
    ```
    which leads to the following traceback
    ```
        pickle.dump(method.subsamplers[0], f)
      File "/home/japarker/projects/PySAGES_main/pysages/methods/core.py", line 50, in __getstate__
        return default_getstate(self)
      File "/home/japarker/projects/PySAGES_main/pysages/methods/core.py", line 235, in default_getstate
        return {key: method.__dict__[key] for key in init_args}, method.kwargs
      File "/home/japarker/projects/PySAGES_main/pysages/methods/core.py", line 235, in <dictcomp>
        return {key: method.__dict__[key] for key in init_args}, method.kwargs
    KeyError: 'center'
    ```

pabloferz:
I think the latest commit should fix the pickling issue.

As for having `executor()` vs `executor` I chose the latter because `MPIPoolExecutor` and other `Excutors` can take initialization arguments so we would either loose that flexibility or would have to also process the extra arguments ourselves. One thing we could do, though is having a function `canonicalize` (or something like this) that takes either instances or class objects and returns only instances (initializing when a class object is passed).

johnaparker:
@pabloferz , @ndtrung81 and I ran some tests on the cluster. In [94e4f19](https://github.com/SSAGESLabs/PySAGES/pull/104/commits/94e4f19b9cd9df9b666a8fa9686c1e20e252f9d3), I just changed the `__getstate__` so that `pickle.load` works correctly. Then the runs are working in parallel.

There is an issue with the final result returned. It stems from the callback; the results of `callbacks[0].data` is an empty list. My thought is the following code will need to be changed:
https://github.com/SSAGESLabs/PySAGES/blob/94e4f19b9cd9df9b666a8fa9686c1e20e252f9d3/pysages/methods/umbrella_integration.py#L119-L128
Here, the callbacks belong to MPI Rank 0, they get passed to MPI worker Rank N, but there is no communication back to Rank 0 on their values. The `HarmonicBiasState` is correctly communicated, because it belongs to `future.result`

One option is to pass `callback_params` to `submit_work`, and have the actual `callback` be stored in the `future`, or something similar to that approach

ndtrung81:
@pabloferz Can I push some changes to this branch? I don't think I have permission at the moment. Thanks!

pabloferz:
You should be able to do it now @ndtrung81.

johnaparker:
Looks good to me as well, I ran Umbrella Integration for the following cases and got expected results:
- Serial execution
- Parallel on 4 GPUs, 1 node
- Parallel on 8 GPUs, 2 nodes

pabloferz:
@johnaparker, @ndtrung81 thank you for looking into this. I'll try to push the rest of the PRs so we can tag a new version soon and then move on to rebase and merge this one.

PR #107, opened 2022-02-14
[CVs] Make `requires_box_unwrapping` true by default

### LINKED ISSUE #105: Collective Variables that need unwrap coordinates.
Here are the CVs that require unwraped coordinates.

Let's discuss if you agree.

> Angle
> DihedralAngle
> Distance
> All the shape CVs

Basically every one of the CVs we have so far.
Maybe we should make that the default behavior and just disable it where it isn't necessary.

pabloferz:
Closes #105

If a user know this is not necessary, it's easy to set it to `False` after the CV is declared but before the sampling method is built.

### DIFF
--- a/pysages/collective_variables/coordinates.py
+++ b/pysages/collective_variables/coordinates.py
@@ -52,10 +52,6 @@ def weighted_barycenter(positions, weights):
 
 
 class Component(AxisCV):
-    def __init__(self, indices, axis):
-        super().__init__(indices, axis)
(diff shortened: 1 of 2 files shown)


pabloferz:
The pre-commit failure can be ignored for now.

InnocentBug:
Let's merge this, for anyone who uses pysages right now.
Thinking of my students.

InnocentBug:
thanks

PR #114, opened 2022-03-08
Docs review (CVs, methods)

ricalessandri:
Went through the CVs again and now through most of the methods. I fixed a number of small things. 
Added an initial description of the FFS method that was missing. 
I will add some comments with other suggestions/things I wasn't sure about.


pysages/collective_variables/shape.py
```diff
@@ -104,7 +104,7 @@ def function(self):
         Returns
         -------
         Callable
-            Function to calculate the Eigenvalue with the specified axis index \
+            Function to calculate the eigenvalue with the specified axis index \
             of the GyrationTensor. \
```
ricalessandri, line 108: --> gyration_tensor?

ricalessandri:
I guess this line:
https://github.com/SSAGESLabs/PySAGES/blob/721cc107eb56d617d31166dffcb359cd75b9ca0b/pysages/collective_variables/core.py#L89
should read `Must be a tuple (size 2) of atoms.` Right?
And similarly for:
https://github.com/SSAGESLabs/PySAGES/blob/721cc107eb56d617d31166dffcb359cd75b9ca0b/pysages/collective_variables/core.py#L109
and
https://github.com/SSAGESLabs/PySAGES/blob/721cc107eb56d617d31166dffcb359cd75b9ca0b/pysages/collective_variables/core.py#L129

pysages/methods/umbrella_integration.py
```diff
@@ -5,11 +5,11 @@
...
-Umbrella integration uses multiple replica placed along a pathway in the free energy
+Umbrella integration uses multiple replicas placed along a pathway in the free energy
 landscape by Harmonic Bias simulations.
 From the statistics of the simulations, the thermodynamic forces along the path are
-determine and integrated to obtain an approximation of the free-energy landscape.
-This class implements the replica simulations and approximates the the free energy.
+determined and integrated to obtain an approximation of the free energy landscape.
+This class implements the replica simulations and approximates the free energy.
 However, the method is not very and accurate it is preferred that more advanced methods
 are used for the analysis of the simulations.
```
ricalessandri, line 14: Are these 2 lines needed? For some systems US may be enough. And eventually, it's up to the user make a wise choice.

pysages/methods/umbrella_integration.py
```diff
@@ -90,11 +90,13 @@ def run(  # pylint: disable=arguments-differ
             Passed to the backend run function as additional user arguments.
 
         * Note:
-            This method does not accepts a user defined callback are not available.
+            This method does not accept a user defined callback are not available.
```
ricalessandri, line 93: unclear to me what is meant here

pysages/methods/ffs.py
```diff
@@ -263,13 +269,15 @@ def basin_sampling(
             print("Restoring basing configuration since system leave basin with cv value:\n")
```
ricalessandri, line 269: ...system leave basin... --> ...system **left** basin... ?
@gustavor101
but I'm not sure since I'm not so familiar with the method.

pysages/methods/ffs.py
```diff
@@ -5,7 +5,11 @@
 """
 Forward Flux Sampling (FFS).
 
-Implementation of the direct version of the Forward Flux Sampling algorithm.
+Implementation of the direct version of the FFS algorithm.
```
ricalessandri, line 8: Eventually probably a reference to the FFS paper(s) may be included somewhere.
(PRL 94, 018104 (2005) and/or J. Chem. Phys. 124, 024102 (2006)?).
I guess either here or maybe even better at:
https://github.com/SSAGESLabs/PySAGES/blob/721cc107eb56d617d31166dffcb359cd75b9ca0b/pysages/methods/ffs.py#L76

pysages/collective_variables/shape.py
InnocentBug: https://en.wikipedia.org/wiki/Gyration_tensor

It has a wikipedia page. Is that googlable enough, or should we write down the equations.

pysages/methods/umbrella_integration.py
InnocentBug: Yes, umbrella integration really isn't that great.
Especially with the approximations that we are making here.
(First order approximation instead of second order.)
We are planning to add better analysis methods later on. For now this should warn users not to expect great results.

pysages/methods/umbrella_integration.py
InnocentBug: Usually all methods accept a callback and the user can write there whatever they want.
UmbrellaIntegration however, does not, because we need a callback for the analysis.

InnocentBug: And we currently haven't implemented anything that allows chaining of callbacks.

pysages/methods/ffs.py
InnocentBug: yeah, I think you are right here @ricalessandri

pysages/methods/ffs.py
InnocentBug: Let's include it now!

InnocentBug (dismissed):
Minor things left from the conversation.
But overall great improvement. Thanks Riccardo.

pysages/methods/umbrella_integration.py
pabloferz:
```suggestion
            This method does not accept a user defined callback.
```

ricalessandri: Gotcha.

pysages/collective_variables/shape.py
ricalessandri: Sorry, my suggestion wasn't clear. I just meant whether we should replace "GyrationTensor" with "gyration_tensor" (the actual name of the function) or actually simply with "gyration tensor". Eventually, I think it should just be "gyration tensor", so I went for that (6e76c1d651812504aa8a9aad7f01b8d0201d177d). If that's OK, go ahead and merge, otherwise I'll change it.

I think there's no need to write down the equations.

pysages/methods/umbrella_integration.py
ricalessandri: Right, right, now I got it. Maybe we could add "more advanced methods (e.g. the Weighted Histogram Analysis Method)". I've added this last bit with the latest commit (6e76c1d651812504aa8a9aad7f01b8d0201d177d). Let me know if you'd like to remove it at this point and I'll amend, otherwise go ahead and merge.

ricalessandri:
Thanks for the feedback. I addressed the comments (6e76c1d651812504aa8a9aad7f01b8d0201d177d). There are a couple conversations that I left open (see my replies above). If you agree with those, please go ahead and merge.

pysages/collective_variables/core.py
```diff
@@ -86,7 +86,7 @@ class TwoPointCV(CollectiveVariable):
     Parameters
     ----------
     indices : list[tuple(int)]
-       Must be a or groups (size 2) of atoms.
+       Must be a tuple (size 2) of atoms.
```
pabloferz, line 89:
```suggestion
       Must be a group (size 2) of atoms.
```

pysages/collective_variables/core.py
```diff
@@ -106,7 +106,7 @@ class ThreePointCV(CollectiveVariable):
     Parameters
     ----------
     indices : list[tuple(int)]
-       Must be a or groups (size 3) of atoms.
+       Must be a tuple (size 3) of atoms.
```
pabloferz, line 109:
```suggestion
       Must be a group (size 3) of atoms.
```

pysages/collective_variables/core.py
```diff
@@ -126,7 +126,7 @@ class FourPointCV(CollectiveVariable):
     Parameters
     ----------
     indices : list[tuple(int)]
-       Must be a or groups (size 4) of atoms.
+       Must be a tuple (size 4) of atoms.
```
pabloferz, line 129:
```suggestion
       Must be a group (size 4) of atoms.
```

pysages/collective_variables/core.py
pabloferz: These are indeed groups. In the line below in each case I tried to indicate that a group is specified as a list or a tuple. If you feel it's not clear enough we can certainly think of a way of improving it.

ricalessandri: Aaah, now I got it. OK I think it's just the extra "or" (`Must be a or groups (size 2)...`) that confused me. Let me commit your suggestions.

pabloferz (approved):

pabloferz:
Thank you @ricalessandri!

PR #115, opened 2022-03-11
Make `run` a module method instead of a class one

pabloferz:
Factored out from #104.


InnocentBug (approved):
Seeing this makes me like Dispatch more and more. 
overloading functions is something I missed with python.

PR #116, opened 2022-03-11
Use composition instead of inheritance for `UmbrellaIntegration`

pabloferz:
Factored out from #104 (rebase once #115 is merged).


InnocentBug (approved):
That looks good to me for now.

I think further down the line, I would like to change it a bit, such that the user can use different smaples.

pabloferz:
Yes, supporting different sampling methods is exactly the idea in the long run.

PR #122, opened 2022-04-12
Remove old nn submodule

pabloferz:
This should also let us drop `fastcore` as dependency.


InnocentBug (approved):
There seem to be no conflicts in removing the module.

It is always good to remove code :)

PR #123, opened 2022-04-14
Update copy to match new jax behavior

pabloferz:
Also, set minimum lower bound for jax to 0.3.5 to make sure new installations work properly.


InnocentBug (approved):

InnocentBug:
Looks good to me and it passes the tests again.

PR #131, opened 2022-05-12
Parallel-bias metadynamics

sivadasetty:


pysages/methods/pbmetad.py
```diff
@@ -0,0 +1,394 @@
...
+
+        return w * cv_switching_probability
+
+    if grid is None:
+        get_grid_index = jit(lambda arg: None)
+        update_grids = jit(lambda *args: (None, None))
+        should_deposit = jit(lambda pred, _: pred)
+
+    else:
+        grid_mesh = compute_mesh(grid) * (grid.size / 2)
```
gustavor101, line 189: grid_mesh variable should have the same value as in [metad.py](https://github.com/SSAGESLabs/PySAGES/blob/main/pysages/methods/metad.py) line 236:
`grid_mesh = (compute_mesh(grid) + 1) * (grid.size / 2) + grid.lower`

PR #132, opened 2022-05-17
Create orientation.py for collective variables

gustavor101:
Collective variables that measure orientation respect a reference using quaternions

### DIFF
--- a/.github/workflows/docker-ci.yml
+++ b/.github/workflows/docker-ci.yml
@@ -121,7 +121,7 @@ jobs:
       - name: Load and run test
         run: |
           docker load --input /tmp/pysages.tar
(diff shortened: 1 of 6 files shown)


pysages/collective_variables/orientation.py
```diff
@@ -0,0 +1,260 @@
...
+    """
+    def __init__(self, indices, axis,ref):
+        super().__init__(indices,axis)
+        self.ref = np.asarray(ref)
+    @property
+    def function(self):
+        return lambda r: eigen(r,self.axis, self.ref)
+
+def eigen(r1,axis,r2):
+        """
```
pabloferz, line 140:
```suggestion
    """
```

PR #133, opened 2022-05-24
MetaD example notebook fixes

ricalessandri:
I was going through the metadynamics example and I noticed a couple of commands that seemed like belonging to a leftover cell (see changes). I removed them (not sure if we instead should change them to something else, in case I can update).


InnocentBug (approved):
LGTM

pabloferz:
I'm going to hijack this PR to add an additional fix to the MetaD notebook (I missed some updates with the new `run` API).

PR #134, opened 2022-05-24
Add orientations-based collective variables 

pabloferz:
Collective variables that measure orientation respect a reference using quaternions.

Supersedes #132


gustavor101 (commented):
1. Barycenter functions imported from coordinates.py are ok.
2. Add weights as input to used weigthed definitions of the CVs.

PR #135, opened 2022-05-26
Provide an `analyze` interface for post-run processing

pabloferz:
Factored out from https://github.com/SSAGESLabs/PySAGES/pull/104.

The idea is to eventually implement analyze functions for every sampling method so the user does not in general require knowledge about the details of each sampling method.

CC. @InnocentBug, @gustavor101, @sivadasetty, @ricalessandri, @ndtrung81

### DIFF
--- a/examples/hoomd-blue/umbrella_integration/integration.py
+++ b/examples/hoomd-blue/umbrella_integration/integration.py
@@ -13,28 +13,30 @@ from pysages.collective_variables import Component
 from pysages.methods import UmbrellaIntegration
 
 
-param1 = {"A": 0.5, "w": 0.2, "p": 2}
+params = {"A": 0.5, "w": 0.2, "p": 2}
 
 
 def generate_context(**kwargs):
     hoomd.context.initialize("")
     context = hoomd.context.SimulationContext()
+
     with context:
         print(f"Operating replica {kwargs.get('replica_num')}")
-        system = hoomd.init.read_gsd("start.gsd")
+        hoomd.init.read_gsd("start.gsd")
 
-        hoomd.md.integrate.nve(group=hoomd.group.all())
-        hoomd.md.integrate.mode_standard(dt=0.01)
+        md.integrate.nve(group=hoomd.group.all())
+        md.integrate.mode_standard(dt=0.01)
 
-        nl = hoomd.md.nlist.cell()
-        dpd = hoomd.md.pair.dpd(r_cut=1, nlist=nl, seed=42, kT=1.0)
+        nl = md.nlist.cell()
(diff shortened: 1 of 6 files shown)


pabloferz:
I'm implementing here the interface only for Umbrella Integration, for the rest of the sampling methods it would be better to do so in separate PRs (one per method) once this PR gets merged.

examples/hoomd-blue/umbrella_integration/integration.py
```diff
@@ -13,28 +13,30 @@
 from pysages.methods import UmbrellaIntegration
 
 
-param1 = {"A": 0.5, "w": 0.2, "p": 2}
+params = {"A": 0.5, "w": 0.2, "p": 2}
 
 
-def generate_context(**kwargs):
+def generate_context(A=0.5, w=0.2, p=2, **kwargs):
```
InnocentBug, line 19: Are we using these anywhere?
We are pulling from `params` right now.

examples/hoomd-blue/umbrella_integration/integration.py
```diff
@@ -13,28 +13,30 @@
...
     context = hoomd.context.SimulationContext()
+
     with context:
         print(f"Operating replica {kwargs.get('replica_num')}")
-        system = hoomd.init.read_gsd("start.gsd")
+        hoomd.init.read_gsd("start.gsd")
 
-        hoomd.md.integrate.nve(group=hoomd.group.all())
-        hoomd.md.integrate.mode_standard(dt=0.01)
+        md.integrate.nve(group=hoomd.group.all())
```
InnocentBug, line 27: why can we omit the `hoomd.` here?
Did we change the imports?

InnocentBug (changes requested):

examples/hoomd-blue/umbrella_integration/integration.py
pabloferz: We don't need these I added them just to realize they weren't necessary, but forgot to remove them.

examples/hoomd-blue/umbrella_integration/integration.py
pabloferz: I changed the imports, but I can revert this change if you think it's clearer to leave the `hoomd.`

InnocentBug: It's OK. I was just a little confused since I missed the import change

examples/hoomd-blue/umbrella_integration/integration.py
InnocentBug: we should remove them then

InnocentBug (approved):

PR #136, opened 2022-05-26
Deprecate collective_variables module in favor of colvars

pabloferz:
And allow sampling methods to skip CVs gradient computation.


pabloferz:
A second option here would be not to deprecate `collective_variables` at all, but to have it as an alias to `colvars` instead.

CC @InnocentBug

pysages/collective_variables/__init__.py
```diff
@@ -5,31 +5,8 @@
...
-from .coordinates import (
-    Component,
-    Distance,
-)
-
-from .utils import (
-    get_periods,
-    wrap,
-)
+warn("Importing `pysages.collective_variables` is deprecated. Import `pysages.colvars` insted")
```
pabloferz, line 12:
```suggestion
warn("Importing `pysages.collective_variables` is deprecated. Import `pysages.colvars` instead")
```

PR #137, opened 2022-06-06
Support for ASE

### LINKED ISSUE #49: [Backends] Support for ASE
ASE would allow us to target a good number of backends all at once (specially QM-MM code bases).

pabloferz:
Initial version of the backend. Needs a lot of testing with different ASE calculators to validate.

Fixes #49


gustavor101 (approved):
Works fine for ase examples

PR #138, opened 2022-06-07
Metadynamics analysis

sivadasetty:
- Added analysis function to `metad.py` for calculating free energy using deposited bias potential.
- Updated metadynamics example notebooks.


pysages/methods/metad.py
```diff
@@ -287,3 +288,24 @@ def sum_of_gaussians(xi, heights, centers, sigmas, periods):
...
+    heights = state.heights
+    centers = state.centers
+    sigmas = state.sigmas
+
+    @jit
+    def metapotential(xs):
+        f = vmap(lambda x: sum_of_gaussians(x, heights, centers, sigmas, P))
+        return f(xs)
+
+    return dict(heights=heights, sigmas=sigmas, metapotential=metapotential)
```
InnocentBug, line 311: Maybe that is me, but what exactly is `heights` and how do  I get the free energy profile from there?

pysages/methods/metad.py
```diff
@@ -287,3 +288,24 @@ def sum_of_gaussians(xi, heights, centers, sigmas, periods):
...
+    """
+    method = result.method
+    state = result.states
+
+    P = get_periods(method.cvs)
+    heights = state.heights
+    centers = state.centers
+    sigmas = state.sigmas
+
+    @jit
```
InnocentBug, line 306: It's nice to compile this with jit, what we are not expecting this to be performance critical, right?

examples/openmm/metad/alanine-dipeptide.py
```diff
@@ -113,11 +113,11 @@ def main(argv=[]):
     callback = MetaDLogger(hills_file, stride) if args.log else None
 
     tic = time.perf_counter()
-    pysages.run(method, generate_simulation, timesteps, callback)
+    run_result = pysages.run(method, generate_simulation, timesteps, callback)
     toc = time.perf_counter()
     print(f"Completed the simulation in {toc - tic:0.4f} seconds.")
```
InnocentBug, line 119: Maybe we can include plotting the results in this example too?

pysages/methods/metad.py
sivadasetty: `heights` in metadynamics refer to the height of the deposited Gaussian potential. It is one of the parameters provided by the user. In standard metadynamics, it is constant that is equal to the user-defined value. For well-tempered metadynamics, it decays exponentially with simulation. It is one of the metric to monitor convergence of the simulations.
To determine free energy, we can use the `metapotential` function. According to standard metadynamics, the free energy is equivalent to the total deposited bias potential. This can be determined using the `metapotential` function. An example is provided in the colab notebook.

pysages/methods/metad.py
sivadasetty: I think `metapotential` evaluation depends on the number of bins used in the CV for computing free energy. So, it can be expensive for 2D or higher dimensions. However, not sure how much performance gain we obtain using `jit` here.

examples/openmm/metad/alanine-dipeptide.py
sivadasetty: This is a good point. I can include steps to plot the results in this file too.

pysages/methods/metad.py
InnocentBug: Thanks for the explanation.
Can we document this for the return of the dictionary with this function?
I think users should be able to find this explanation there.

pysages/methods/metad.py
InnocentBug: Yeah, I think it is fine. @pabloferz is there a drawback that we should consider here for users?

examples/openmm/metad/alanine-dipeptide.py
InnocentBug: Thanks!

pysages/methods/metad.py
sivadasetty: Sure, I can add that. I think we don't need to return `sigmas` here. So, I ll also edit that. Thanks!

pysages/methods/metad.py
pabloferz: I don't think it will hurt even if this not performance critical, and I don't think this will be problematic for the end-user, they should be able to plot or write to file `DeviceArray`s in the same way they do with `nd.array`s.

examples/openmm/metad/alanine-dipeptide.py
sivadasetty: Done.

InnocentBug (approved):

InnocentBug (dismissed):

pysages/methods/metad.py
```diff
@@ -288,3 +288,45 @@ def sum_of_gaussians(xi, heights, centers, sigmas, periods):
     """
     delta_x = wrap(xi - centers, periods)
     return gaussian(heights, sigmas, delta_x).sum()
+
+
+@dispatch
+def analyze(result: Result[Metadynamics]):
+    """
+    Helps in calculating the free energy from the final state of `Metadynamics` run.
```
pabloferz, line 296:
```suggestion
    Helper for calculating the free energy from the final state of a `Metadynamics` run.
```

pysages/methods/metad.py
```diff
@@ -288,3 +288,45 @@ def sum_of_gaussians(xi, heights, centers, sigmas, periods):
...
+        heights:
+            Height of the Gaussian bias potential during the simulation.
+
+        metapotential:
+            Function that takes user-defined CV range to compute the deposited bias potential.
+            For standard metadynamics, the free energy along user-defined CV range is the same
+            as the metapotential(cv). In the case of well-tempered metadynamics, the the
+            free energy along user-defined CV range is equal to
+            (T + deltaT)/deltaT x metapotential(cv), where T is the simulation temperature
+            and deltaT is the user-defined paramter in well-tempered metadynamics.
```
pabloferz, line 316:
```suggestion
            Maps a user-provided array of CV values to the corresponding deposited bias potential.
            For standard metadynamics, the free energy along user-provided CV range is the same
            as `metapotential(cv)`. In the case of well-tempered metadynamics, the free energy is
            equal to `(T + deltaT) / deltaT * metapotential(cv)`, where `T` is the simulation temperature
            and `deltaT` is the user-defined parameter in well-tempered metadynamics.
```

pabloferz (changes requested):

pysages/methods/metad.py
```diff
@@ -288,3 +288,44 @@ def sum_of_gaussians(xi, heights, centers, sigmas, periods):
...
+
+        heights:
+            Height of the Gaussian bias potential during the simulation.
+
+        metapotential:
+            Maps a user-provided array of CV values to the corresponding deposited bias potential.
+            For standard metadynamics, the free energy along user-provided CV range is the same
+            as `metapotential(cv)`. In the case of well-tempered metadynamics, the free energy is
+            equal to `(T + deltaT) / deltaT * metapotential(cv)`, where `T` is the simulation temperature
+            and `deltaT` is the user-defined parameter in well-tempered metadynamics.
```
pabloferz, line 315:
```suggestion
            equal to `(T + deltaT) / deltaT * metapotential(cv)`, where `T` is the simulation
            temperature and `deltaT` is the user-defined parameter in well-tempered
            metadynamics.
```

pabloferz (approved):

PR #139, opened 2022-06-07
Add some details to making working with PySAGES easier

InnocentBug:


pysages/colvars/core.py
```diff
@@ -9,15 +9,16 @@
...
 
 from jax import grad as jax_grad, jit, numpy as np
+import numpy
 
 from pysages.utils import JaxArray, dispatch
 
 
 UInt32 = np.uint32
-Indices = Union[int, range]
+Indices = Union[numpy.intp, numpy.int, int, range]
```
pabloferz, line 21: Isn't `numpy.intp` enough here?

pabloferz, line 21: Isn't `numpy.intp` enough here?

InnocentBug: yes, numpy.intp is enough.
But I wasn't sure what else ppl might throw at PySAGES.
Could be a custom created array with the indices.

Does it hurt here?

InnocentBug: see above

pabloferz: I'm asking because `numpy.int` got deprecated in numpy `1.20` and `numpy.intp` should point to the default precision. I think what you might want is `numpy.intc` and `numpy,int_` instead.

InnocentBug: oh, ok I wasn't aware of that.
Let's add all of them instead of the deprecated one?

pysages/colvars/core.py
```diff
@@ -285,7 +286,7 @@ def _is_group(obj):
 
 
 @dispatch
-def _group_size(obj: int):  # pylint:disable=unused-argument
+def _group_size(obj: Union[numpy.intp, numpy.int, int]):  # pylint:disable=unused-argument
```
pabloferz, line 289:
```suggestion
def _group_size(obj: Union[numpy.intp, numpy.intc, numpy.int_, int]):  # pylint:disable=unused-argument
```

pabloferz (approved):

PR #140, opened 2022-06-07
Add a post function to run after `backend.run` to finish up or store final results

InnocentBug:


InnocentBug:
There is a slight problem with hoomd-blue using this.

The feature that hoomd reports the last executed python line gets confused and just prints out the entire script.

Not sure if we should do something about that, document it, or leave it as is.

pysages/methods/umbrella_integration.py
```diff
@@ -102,6 +103,13 @@ def run(  # pylint: disable=arguments-differ
     kwargs:
         Passed to the backend run function as additional user arguments.
 
+    post_run_action: Optional[Callable] = None
+        Callable function that enables actions after the run execution of PySAGES.
+        Actions are executed inside the generated context.
+        Application for this is writing a final configuration file etc.
+        This function get `context_args` unpacked for additional user
+        arguments just like `context_generator`.
```
pabloferz, line 111:
```suggestion
        Example uses for this include writing a final configuration file.
        This function gets `context_args` unpacked just like `context_generator`.
```

pabloferz (approved):

PR #142, opened 2022-06-14
Add gradient-based objective functions

pabloferz:
I intended to add these during the ml rewrite. These are not being used yet, but are prep work for some of the methods missing such as SIRENs sampling.


gustavor101 (approved):
Changes works well with the test for analyzer for FUNN and ABF

PR #147, opened 2022-06-21
Ensure all sampling methods are picklable

pabloferz:
Factored out from #104.

### DIFF
--- a/pysages/grids.py
+++ b/pysages/grids.py
@@ -26,6 +26,10 @@ class Chebyshev(GridType):
     pass
 
 
+class NoGrid:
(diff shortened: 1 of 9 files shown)


pabloferz:
`Grid` is a bit hard to make pickleable, that' why I get the grid info so we are able to reconstruct in `__setstate__`. I had tested before and it was working within the `parallel` branch. Let me check why is it failing here.

pabloferz (commented):
Should be ready now @InnocentBug

InnocentBug (approved):

PR #149, opened 2022-06-24
Reimplement parallelization API

InnocentBug:
This reimplements @johnaparker's parallel API (supersedes #104)


pabloferz:
Fixes #54

InnocentBug:
I successfully ran OpenMM and Hoomd-blue examples in berlin.
If the test pass, we can go ahead with this.

.gitignore
```diff
@@ -6,4 +6,11 @@ docs/build
...
-examples/*/*gsd
\ No newline at end of file
+examples/*/*gsd
+examples/*/*/*pdf
+examples/*/*/*dat
+examples/*/*/*gsd
+dy_periodic_sirens_sobolev_fit.pdf
+examples/*/.ipynb_checkpoints/
+y_mlp_fit.pdf
+y_periodic_sirens_sobolev_fit.pdf
```
pabloferz, line 16: I would use `*.pdf` instead.

InnocentBug: sure, NP

pabloferz:
I can't figure out why the hoomd example is not working here. I wonder if we should remove that form our CI for now and merge and figure out the issue later.

pabloferz (dismissed):

pabloferz (approved):

PR #154, opened 2022-07-05
Introduce restraints

pabloferz:
Often it's necessary to limit systems to sample only within a collective variables subdomain. This PR adds the mechanism to do so for the sampling methods for which this might be needed,


gustavor101 (dismissed):
Much more clear definition now of how to apply restraints in gridded methods

InnocentBug (dismissed):

PR #156, opened 2022-07-13
Tweak docker-ci workflow

pabloferz:
Hopefully this will fix the random failures when running with MPI in docker .


pabloferz:
It looks like the unsuccessful runs where due to OOM faillures, reducing the number of concurrent processes seems to reduce the RAM usage enough.

PR #159, opened 2022-07-17
[Methods] Add new SpectralABF method

pabloferz:


gustavor101 (dismissed):
New method works well in external test. Adding an example in the openmm folder will be great. A consistent dictionary across analyze in all methods should be great.

PR #160, opened 2022-07-19
[Methods] Add CFF method

pabloferz:


InnocentBug (changes requested):
Can we add the Butane example also as automated test?

InnocentBug:
closes #52

InnocentBug (approved):

PR #161, opened 2022-07-19
Make safer to iterate over futures

pabloferz:


pysages/methods/umbrella_integration.py
```diff
@@ -135,7 +136,7 @@ def submit_work(executor, method, context_args, callback):
             local_context_args["replica_num"] = rep
             callback = method.histograms[rep]
             futures.append(submit_work(ex, submethod, local_context_args, callback))
-        results = [future.result() for future in futures]
+        results = [future.result() for future in as_completed(futures)]
```
InnocentBug, line 139: I don't think this is what we want.
As completed returns the futures when they are done, but we want the results in order.
It doesn't look to me that as_completed guarantees that.

Umbrella integration needs the results in order. For the default one, that might be OK, but could also lead to confusion.

Maybe instead, we call executor.shutdown(wait=True, cancel=False) instead.
That should guarantee that all futures are done.

I can't find the documentation if the __exit__ function of executors.
I would assume that it calls cancel somehow.
If that would be the case, we would just have to move the future.result block out of the with block.


We have to wait for all to finish anyways before we can start analysis. So not worth getting some results earlier just because the finish faster.

InnocentBug (changes requested):

pysages/methods/umbrella_integration.py
pabloferz: Ah, right. This should not be needed. When used as a context manager the executor should call `shutdown` at the end.

InnocentBug:
well, actually we should reopen it.
We are not leaving the with block at the moment before we call result.

Do you want me to do it in a new PR?

PR #163, opened 2022-07-19
Make Distance a "COG Distance" (handles also groups of points)

ricalessandri:
Something we discussed with @pabloferz and found that it may make for a more intuitive and flexible behavior for `Distance`.

Basically, we make the `Distance` CV a "barycenter distance" so that it can handle both distances between single points but also between group of points (by computing the distance between the barycenter aka center_of_geometry of the groups of atoms).

### DIFF
--- a/pysages/colvars/coordinates.py
+++ b/pysages/colvars/coordinates.py
@@ -90,7 +90,7 @@ class Distance(TwoPointCV):
 def distance(p1, p2):
     """
     Returns the distance between two points in space or
-    between the barycenter of two groups of points in space.
+    between the barycenters of two groups of points in space.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


InnocentBug (commented):
LGTM

pysages/colvars/coordinates.py
```diff
@@ -89,18 +89,22 @@ def function(self):
 
 def distance(p1, p2):
     """
-    Returns the distance between two points in space.
+    Returns the distance between two points in space or
+    between the barycenter of two groups of points in space.
```
pabloferz, line 93:
```suggestion
    between the barycenters of two groups of points in space.
```

pabloferz (approved):

pabloferz (approved):

PR #164, opened 2022-07-19
Allow unbiased simulation to collect CVs

InnocentBug:
Not the most elegant implementation since we are still adding zeros in the simulations.
But since should only be test simulations, I think it is good to go.


InnocentBug:
Maybe down the line we should improve this implementation.

I can see people using PySAGES just for this.
Implementing their own CV to measure an observable in their simulation and log it.

It's easy to do and depending on your backend (Hoomd easy on cpu, OpenMM Idk: hard) it is worthwhile.
And if ppl use it like this, we should circumvent the addition of the zero force altogether.

pabloferz:
Can we call it `Unbiased`?

Regarding avoiding adding the forces, I think we can do it by setting the bias to `None`, and changing the backends code to conditionally add them if `state.bias is not None`.

InnocentBug:
Great suggestions, I implemented both.

examples/hoomd-blue/unbiased/.gitignore
```diff
@@ -0,0 +1 @@
+start.gsd
```
pabloferz, line 1: Isn't this covered by our root `.gitignore`?

InnocentBug: I think it is.
I copied it from harmonic_bias example.
Does it hurt?

pabloferz: It doesn't, it's just my picky side speaking. I can ignore it for now.

pabloferz:
I added a commit to avoid computing the Jacobians as well.

pabloferz (approved):

PR #169, opened 2022-07-21
Specialize distance when using groups

pabloferz:


InnocentBug:
I tried in the notebook.
And it didn't solve the issue for me :(

pabloferz:
Let me look at it again. Are we sure that #163 introduced the change in behaviour?

pabloferz:
I added a test comparing the result when using groups vs when not and they seem to give the same result and it seems to be working. What are the differences you are seeing with this PR?

InnocentBug:
Ok, I double checked. Now it works.

InnocentBug (approved):

PR #170, opened 2022-07-28
Small fixes of the openmm examples

ricalessandri:
Some links of recent (and less recent) addition to the OpenMM script example folder where missing links in the README file.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -34,8 +34,11 @@ Examples for Methods using OpenMM can be found in the subfolder [openmm](openmm)
 
 ### Script examples
 
-* ABF with Alanine Dipeptide: [ABF](openmm/abf)
+* ABF with Alanine Dipeptide: [ABF](openmm/abf/alanine-dipeptide_openmm.py)
 * Forward flux sampling with Alanine Dipeptide: [FFS](openmm/forward_flux_sampling.py)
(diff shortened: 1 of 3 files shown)


ricalessandri:
Also, I don't think the lines:
```
mkdir /content/cff
cd /content/cff
```
in the following files:
* [Butane_CFF.md](https://github.com/ricalessandri/PySAGES/blob/example-links/examples/hoomd-blue/cff/Butane_CFF.md)
* [Butane_CFF.ipynb](https://github.com/ricalessandri/PySAGES/blob/example-links/examples/hoomd-blue/cff/Butane_CFF.ipynb)

are supposed to be there, or are they? 
I'm raising this point because if one runs the notebook on some machine where they don't have permissions to write in / - like at the IRIS Berlin machines - those line won't work.

InnocentBug:
Yes, these lines are meant for google colab only.
But that should be a huge deal. Having them fail should be fine. 
i think the setup section uses the /content/ explicitly too. 
we could move these line into the setup section. So, locally you just don’t run these anyways.

InnocentBug (approved):

ricalessandri:
Gotcha. I like @InnocentBug 's idea to move those lines to the setup section of the notebooks. I made this into an issue (#175 ) and I'm merging the current PR.

PR #171, opened 2022-07-28
Update ann examples

ricalessandri:
Update the ANN example for butane to use the analyze ANN method that was recently (45df887547f8f9332dde0177fbff54d53c149e51) implemented.


ricalessandri:
There's now some inhomogeneity in the way the examples are organized in [`examples/hoomd-blue`](https://github.com/SSAGESLabs/PySAGES/tree/main/examples/hoomd-blue). For example, CFF has the notebook inside the [`cff/` subfolder](https://github.com/SSAGESLabs/PySAGES/tree/main/examples/hoomd-blue/cff) but FUNN doesn't. I was editing this ANN example to follow CFF but wanted to know what you guys @pabloferz @InnocentBug think first.

pabloferz (approved):
Yes, we should reorganize in subfolders per method.

ricalessandri:
I added also a script - `butane_ann.py` (9631480a07581a46fbf00f2484e0b386df864176). It also write the dihedral vs free energy to a `.dat` file for eventual after-run re-plotting (I find that useful).

InnocentBug:
Writing the results in a .dat file is one way of doing this.
I think the more python way of doing it, would be to pickle the results.
But either is fine.

InnocentBug (approved):

ricalessandri:
I'll merge now this and create an issue about later (i.e., after the workshop) switching from this .dat file saving to pickling.

PR #177, opened 2022-08-01
Enable custom bias classes for umbrella integration

InnocentBug:


pysages/methods/harmonic_bias.py
```diff
@@ -59,15 +59,13 @@ def __init__(self, cvs, kspring, center, **kwargs):
...
         """
-        super().__init__(cvs, **kwargs)
+        super().__init__(cvs, center, **kwargs)
         self.cv_dimension = len(cvs)
         self.kspring = kspring
-        self.center = center
 
     def __getstate__(self):
-        state, kwargs = default_getstate(self)
+        state, kwargs = super().__getstate__()
```
InnocentBug, line 67: In general, shouldn't we implement `default_getstate` as the implementation of the general parent class?
Here of `SamplingMethod` and call `super().__getstate__()` in the inheriting classes.
It should work, that the parent class finds the correct `__special_args__` of the inherited classes and argument structure of the constructors, right?

pabloferz: Yes. That should work as well. I believe I thought that we might want to do special pickling for CVs or other classes as well and that we could reuse the same code structure and base functions. But if there's reason to believe that won't be the case, we can replace this.

InnocentBug: Ok, I can take care of this on Monday

InnocentBug:
Locally, I cannot replicate the failiong MPI test.
I have to look in more detail later.

pabloferz (approved):

PR #182, opened 2022-08-02
Implementation of the string method

InnocentBug:
Using umbrella integration as a sub-method for string method.

For now, I am planning to use scipy interpolation for this.
No need to use GPU communicated fast splines on the CV space.
We can add that later when needed (and implemented in JAX.scipy).


InnocentBug:
I finally could test the string implementation with OpenMM.
And that allowed me to test with Alanine Dipeptide.

I think the results look rather promising.
The path is obviously not converged, but it looks reasonable to me.
@ricalessandri @sivadasetty @pabloferz do the results look reasonable for what you would expect for Alanin-Dipeptide?

InnocentBug:
[path_hist.bak.pdf](https://github.com/SSAGESLabs/PySAGES/files/9419567/path_hist.bak.pdf)
[energy_bak.pdf](https://github.com/SSAGESLabs/PySAGES/files/9419568/energy_bak.pdf)

Sorry missing files:

- The analyzed energy profile and convergence criterion for each data point.
- The path as it evolves during the string algorithm.

InnocentBug:
This implementation relies a bit on `post_run_action` to initialize the simulations after each iteration with the last simulation snapshot.

I think that is OK for now.
Ideally, we wouldn't need to write this to disk.
As we implement the restart capability, we will have a way to initialize a simulation from a snapshot as taken in the simulation.
This mechanism should also work to help the string to set the simulation states to a given snapshot - circumventing the writing of configurations to disk.

I can open an issue for that if we want.

InnocentBug:
[energy.pdf](https://github.com/SSAGESLabs/PySAGES/files/9457039/energy.pdf)
[path_hist.pdf](https://github.com/SSAGESLabs/PySAGES/files/9457040/path_hist.pdf)
Updated results from Alanine dipeptide.
500 iterations; it looks good and the free-energy barrier goes down.
So I am pretty sure it is doing the right thing now.

InnocentBug:
And comparing it to this
https://www.plumed.org/doc-v2.7/user-doc/html/marvel-2.html
it looks pretty good.
I mixed up the angles and direction, but other than that it seems like we are finding the right path.
Plus the free energy-barrier seems to be in the right order of magnitude.

Ready for review.

pysages/methods/umbrella_integration.py
```diff
@@ -165,12 +165,11 @@ def submit_work(executor, method, context_args, callback):
         )
 
     futures = []
-    with executor as ex:
-        for rep, submethod in enumerate(method.submethods):
-            local_context_args = deepcopy(context_args)
-            local_context_args["replica_num"] = rep
-            callback = method.histograms[rep]
-            futures.append(submit_work(ex, submethod, local_context_args, callback))
+    for rep, submethod in enumerate(method.submethods):
```
pabloferz, line 168: Why is this removing the context manager?

InnocentBug: Because the context manager calls `shutdown` when you are leaving.
And an MPI pool cannot be restarted after `shutdown` was called.

Which is fine for Umbrella integration, where we use the pool once.
But for the string you want to run multiple umbrella integrations and that requires using the MPI pool multiple times.

pabloferz: Got it. I'd still want to have this probably run conditionally, but we can figure out a way to do so later.

InnocentBug: I also checked and calling `result()` on a future ensures that it is done with its work.
So we shouldn't have sync issues.

InnocentBug: Yeah, maybe we add something that optionally shutdowns the communicator at the end of a run.
It's probably better to explicitly release the resources.

pysages/methods/improved_string.py
```diff
@@ -0,0 +1,295 @@
...
+    spacing = np.cumsum(spacing)
+    assert abs(spacing[0]) < 1e-6
+    assert abs(spacing[-1] - 1) < 1e-6
+    spacing[0] = 0.0
+    spacing[-1] = 1.0
+
+    return spacing
+
+
+class ImprovedString(SamplingMethod):
```
pabloferz, line 51: Even if we describe it as Improved String Method in the docs could we call this just `StringMethod`? Or maybe find a different name for it?

pysages/methods/improved_string.py
```diff
@@ -0,0 +1,295 @@
...
+    stringsteps: Union[int, float],
+    context_args: Optional[dict] = None,
+    post_run_action: Optional[Callable] = None,
+    executor=SerialExecutor(),
+    **kwargs
+):
+    """
+    Implementation of the improved string method.
+    The
+    order (ignoring second order terms with covariance matrix) as described in
```
pabloferz, line 183:
```suggestion
    The order (ignoring second order terms with covariance matrix) as described in
```

pysages/methods/improved_string.py
InnocentBug: I thought Improved String Method was a more or less standing term for a specific implementation of the general string method.
https://www.researchgate.net/profile/Eric-Vanden-Eijnden/publication/6351826_Simplified_and_Improved_String_Method_for_Computing_the_Minimum_Energy_Paths_in_Barrier_Crossing_Events/links/5523eeab0cf2c815e073df2f/Simplified-and-Improved-String-Method-for-Computing-the-Minimum-Energy-Paths-in-Barrier-Crossing-Events.pdf

In particular, the improved string method uses a spline to interpolate the path instead of using derivatives of the path without interpolation.

If we want to make it more specific. I could see `SplineString` a good alternative word.
But I don't know if anyone else has ever called the string method that.

pysages/methods/umbrella_integration.py
InnocentBug: @pabloferz are you OK with this?

pysages/methods/improved_string.py
pabloferz: When I read that reference the `improved` doesn't really make me think that that is part of the name of the new method. I'd like to be more specific, and `SplineStringMethod` or `SplineString` are good names in my opinion. I'd still use in the docs and docstrings something like *This is an implementation of the improved string method proposed/described in...*

pysages/methods/umbrella_integration.py
pabloferz: Yes, this is fine for now. We can find a way to be explicit about calling `shutdown` later.

pysages/methods/improved_string.py
InnocentBug: Yeah, I know it is bad naming.
(Never name anything, "new", "improved", "novel" etc. I hate that in paper names.)

This sounds good to me.
I will change the names accordingly.
Of the class and in the doc.

pysages/methods/umbrella_integration.py
InnocentBug: It is actually quite simple to implement the shutdown.
So I did this here: https://github.com/SSAGESLabs/PySAGES/pull/182/commits/c2dfb817ee43a096785daec1342a4a9bc508f186
Pending tests, I think that should do it for us.

InnocentBug: Let me know what you think.

pysages/methods/improved_string.py
```diff
@@ -0,0 +1,301 @@
...
+            executor,
+            **umbrella_kwargs
+        )
+
+        new_xi = []
+        for i in range(len(umbrella_result.callbacks)):
+            sampled_xi = umbrella_result.callbacks[i].get_means()
+            old_xi = method.umbrella_sampler.submethods[i].center
+            direction = (sampled_xi - old_xi) / method.metric(sampled_xi, old_xi)
+            new_xi.append(np.asarray(old_xi) + np.asarray(method.alpha) * np.asarray(direction))
```
pabloferz, line 248: Can we call `np.asarray` over alpha during `__init__` and just use `method.alpha` here?

pysages/methods/umbrella_integration.py
pabloferz: That works for me!

pysages/methods/improved_string.py
InnocentBug: sure: https://github.com/SSAGESLabs/PySAGES/pull/182/commits/6f51494602de49d506a61be11edcc9eea7871136

pysages/methods/improved_string.py
```diff
@@ -0,0 +1,300 @@
...
+            if i not in method.freeze_idx:
+                method.umbrella_sampler.histograms[i].reset()
+        umbrella_result = pysages.run(
+            method.umbrella_sampler,
+            context_generator,
+            timesteps,
+            context_args,
+            post_run_action,
+            executor,
+            executor_shutdown=False**kwargs,
```
pabloferz, line 239:
```suggestion
            executor_shutdown=False.
            **kwargs,
```

pysages/methods/spline_string.py
```diff
@@ -0,0 +1,297 @@
...
+
+"""
+Spline String method aka Improved String method.
+
+The improved string methods finds the local minimum free-energy path (MFEP) trough a given landscape.
+It uses umbrella integration to advance the discretized replicas in the CV space.
+A spline interpolation helps to (equally) space the CV points along the string.
+The process converges if the MFEP is found.
+The final free-energy profile is calculated the same way as in UmbrellaIntegration.
+We aim to implement this: `Weinan, E., et. al. J. Chem. Phys. 126.16 (2007): 164103 <https://www.researchgate.net/profile/Eric-Vanden-Eijnden/publication/6351826_Simplified_and_Improved_String_Method_for_Computing_the_Minimum_Energy_Paths_in_Barrier_Crossing_Events/links/5523eeab0cf2c815e073df2f/Simplified-and-Improved-String-Method-for-Computing-the-Minimum-Energy-Paths-in-Barrier-Crossing-Events.pdf>`_.
```
pabloferz, line 13:
```suggestion
We aim to implement this: `Weinan, E., et. al. J. Chem. Phys. 126.16 (2007): 164103 <https://doi.org/10.1063/1.2720838>`_.
```

or did you wanted to provide a direct link to the PDF for everyone to access? I think Google Scholar provides this link as well, so I'm no sure what to use here.

docs/source/module-pysages-methods-string.rst
```diff
@@ -0,0 +1,14 @@
+Improved String Method
+----------------------
```
pabloferz, line 2:
```suggestion
Spline String Method
--------------------
```

pysages/methods/spline_string.py
```diff
@@ -0,0 +1,297 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2020-2021: PySAGES contributors
+# See LICENSE.md and CONTRIBUTORS.md at https://github.com/SSAGESLabs/PySAGES
+
+"""
+Spline String method aka Improved String method.
```
pabloferz, line 6:
```suggestion
Spline String method (aka Improved String method).
```

pabloferz (approved):

PR #196, opened 2022-08-15
use eigh instead of eigvals

### LINKED ISSUE #195: Acylindricity and Asphericity CV doesn't work on GPU
`NotImplementedError: Nonsymmetric eigendecomposition is only implemented on the CPU backend. If your matrix is symmetric or Hermitian, you should use eigh instead.`

InnocentBug:
close #195

### DIFF
--- a/pysages/colvars/shape.py
+++ b/pysages/colvars/shape.py
@@ -173,7 +173,7 @@ def principal_moments(positions):
     DeviceArray
         Eigenvalues of the gyration tensor
     """
-    return linalg.eigvals(gyration_tensor(positions))
+    return linalg.eigh(gyration_tensor(positions))[0]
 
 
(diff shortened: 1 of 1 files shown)


pysages/colvars/shape.py
```diff
@@ -173,7 +173,7 @@ def principal_moments(positions):
     DeviceArray
         Eigenvalues of the gyration tensor
     """
-    return linalg.eigvals(gyration_tensor(positions))
+    return linalg.eigh(gyration_tensor(positions))[0]
```
pabloferz, line 176:
```suggestion
    return linalg.eigvalsh(gyration_tensor(positions))
```

pabloferz (approved):

InnocentBug:
`/trunk merge`
;)

PR #201, opened 2022-08-24
Enable meaningful post action triggers for OpenMM

InnocentBug:
Store the generated context into `context_args`.

This allows direct access to the simulation context in the `post_run_actions` to write final simulation files with OpenMM.
Hoomd-blue doesn't need it, since you can automatically operate on the currently activated context, but OpenMM seems to require direct access to the `simulation` object.

With the PR the following is possible for OpenMM.

```python
def post_run_action(**kwargs):
  kwargs.get("context").saveState("final.xml")
```

The equivalent for hoomd-blue is:
```python
def post_run_action(**kwargs):
      hoomd.dump.gsd(filename="final.gsd", overwrite=True, period=None, group=hoomd.group.all())
```
And it doesn't need direct access to the simulation context.

This makes restarting and keeping a checkpoint at the end possible.

### DIFF
--- a/examples/openmm/abf/alanine-dipeptide_openmm.py
+++ b/examples/openmm/abf/alanine-dipeptide_openmm.py
@@ -105,13 +105,16 @@ def save_energy_forces(result):
     numpy.savetxt("Forces.csv", numpy.hstack([grid, forces.reshape(-1, grid.shape[1])]))
 
 
+def post_run_action(**kwargs):
+    kwargs.get("context").saveState("final.xml")
+
 # %%
 def main():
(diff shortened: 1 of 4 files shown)


InnocentBug:
Sorry, I missed this with original introduction of `post_run_action`.

The `context_args` should never be pickled or anything after, so it should be safe to store it there for access.
There are other options of what we could do to enable this if there is an issue we storing the context in context_args.

pabloferz (approved):

InnocentBug:
At some point we should probably document this in a way.

InnocentBug:
Will merge after test pass.

Issue #189, opened 2022-08-11
Segmentation fault error with Hoomd-blue examples 


angelaabongwa:
I believe there's a backend issue with either the extension module or the python interpreter. When I run any example in hoomd-blue, I get an error: Segmentation fault (core dumped)

pabloferz:
Thanks for the bug report Angela. Does this happen at the end of the run or in the middle of the simulation? If it happens at the end, are you getting the results of the simulation?

I suspect this might have been introduced on some of the most recent changes in the hoomd-dlext repo.

angelaabongwa:
It happens at the end and the output files (energy, histogram) are the updated accordingly.

InnocentBug:
So your simulations finish correctly and only at the very end you experience a crash from the python interpreter?

Do you happen to run examples with MPI support?

InnocentBug:
There are known issues with segfault with hoomd-blue.

-  If you initialize `hoomd.context.initialize()` inside an existing context, segfaults are expected.
   Do not do this in your `generate_context` function.
   ```python
   context = hoomd.context.SimulationContext()
   with context:
      hoomd.context.initialize()
      ...
   ```
   Instead do this:
   ```
   hoomd.context.initialize()
   context = hoomd.context.SimulationContext()
   with context:
      ...
   ```
-  At the end of simulations, XLA might try to release resources that are already released before.
   This is a known issue. We are working on a solution, but it is not priority right now, since all simulations are correctly executed.
   
The reported issue seems to be the latter, so I am going to close the issue for now.

PR #206, opened 2022-09-10
Ensure that the context manager is exited

pabloferz:
For some reason the SystemView context manager was not always being exited. This seems to work and it should be safe as we are now returning copies of the arrays in `take_snapshot`.

As far as I can tell this fixes the segfault reported in #189 (both jax and hoomd-dlext were trying to free the arrays from the initial snapshot).

### DIFF
--- a/pysages/backends/hoomd.py
+++ b/pysages/backends/hoomd.py
@@ -107,11 +107,11 @@ def is_on_gpu(context):
 def take_snapshot(wrapped_context, location=default_location()):
     context = wrapped_context.context
     sysview = wrapped_context.view
-    positions = asarray(positions_types(sysview, location, AccessMode.Read))
-    vel_mass = asarray(velocities_masses(sysview, location, AccessMode.Read))
-    forces = asarray(net_forces(sysview, location, AccessMode.ReadWrite))
-    ids = asarray(rtags(sysview, location, AccessMode.Read))
-    imgs = asarray(images(sysview, location, AccessMode.Read))
+    positions = copy(asarray(positions_types(sysview, location, AccessMode.Read)))
+    vel_mass = copy(asarray(velocities_masses(sysview, location, AccessMode.Read)))
+    forces = copy(asarray(net_forces(sysview, location, AccessMode.ReadWrite)))
+    ids = copy(asarray(rtags(sysview, location, AccessMode.Read)))
+    imgs = copy(asarray(images(sysview, location, AccessMode.Read)))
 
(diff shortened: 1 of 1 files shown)


pysages/backends/hoomd.py
```diff
@@ -107,11 +107,11 @@ def is_on_gpu(context):
 def take_snapshot(wrapped_context, location=default_location()):
     context = wrapped_context.context
     sysview = wrapped_context.view
-    positions = asarray(positions_types(sysview, location, AccessMode.Read))
-    vel_mass = asarray(velocities_masses(sysview, location, AccessMode.Read))
-    forces = asarray(net_forces(sysview, location, AccessMode.ReadWrite))
-    ids = asarray(rtags(sysview, location, AccessMode.Read))
-    imgs = asarray(images(sysview, location, AccessMode.Read))
+    positions = copy(asarray(positions_types(sysview, location, AccessMode.Read)))
```
InnocentBug, line 110: Is that OK.
What are we copying here.
Just the pointer to the data?
The context manager for the data?
Or a deep copy?

pysages/backends/hoomd.py
```diff
@@ -197,20 +197,20 @@ def bind(
...
+    wrapped_context.view = sysview
+    wrapped_context.run = hoomd.run
     helpers, restore, bias = build_helpers(context, sampling_method)
 
-    with SystemView(context.system_definition) as sysview:
-        wrapped_context.view = sysview
-        wrapped_context.run = hoomd.run
-
+    with sysview:
         snapshot = take_snapshot(wrapped_context)
```
InnocentBug, line 206: So, the only function we call inside the context manager is the snapshot.
But don't we lose the context manager access for the snapshot once we leave the sysview context?

Seems to me like we use a dangling pointer here.

pysages/backends/hoomd.py
pabloferz: This copies the whole array. But this is only called once at the very beginning of the run (this is different than the one called internally in `Sampler`). But I think I can get it to work without calling `copy` here.

pabloferz: Solving this differently is going to be trickier than I thought, but this should work fine for now. I'd say we move ahead and find a better way to fix it later.

pysages/backends/hoomd.py
pabloferz: We won't because, we would be returning copies of all arrays.

InnocentBug (approved):
Ok, that is fine with me for now.

PR #212, opened 2022-09-16
Fix NACL test with new FF URL

### LINKED ISSUE #209: Two files are almost the same.
`examples/openmm/metad/nacl/nacl.py`
and 
`examples/openmm/metad/nacl/nacl-displacement.py`
are almost the same file.

Can maybe one of them import stuff from the other one.
Maintance seems hard here.

InnocentBug:
Also consolidate the two NACL files.
Close #209


pabloferz (approved):

ricalessandri (commented):
LGTM. Is the test failing, though?

Also, should we get rid of these commented-out lines:
https://github.com/SSAGESLabs/PySAGES/blob/b33aae291e1c8aba82e751abf59c7c32e119d2b3/examples/openmm/metad/nacl/nacl.py#L85-L86

InnocentBug:
Sure, let's get rid of these lines

InnocentBug:
Yes, it should work now.
And it works locally on my machine.

The error, I am getting in the test is HTTP 400, while before we got 404 for the missing file.
I am not sure, why we are getting 400 now. Seems like an odd error for me.
I am ready to locally stores this input file and not download it from the web everytime.

What do you think?

pabloferz:
HTTP 400 is a bad request, but this works locally, so we can ignore this for now. Can we update the corresponding Colab notebooks? I just remembered those also were written with the previous URL

InnocentBug:
Ok, I fixed the notebook now as well.

(There were more issues)

Howver @pabloferz , I also noticed that OpenMM doesn't work in the notebooks right now.
Is OpenMM missing from the notebook env now?

InnocentBug:
Also, I would really like for the test to not fail anymore.
I am trying to setup merge queues for us and it won't merge if we have failing tests.

That's where my locally stored forcefield idea is coming from.

InnocentBug:
Ahh, it doesn't work.

pabloferz (approved):

PR #216, opened 2022-09-23
Use numpy numpy arrays in results, instead of JAX.numpy arrays

InnocentBug:
For the final result, we should prefer using normal numpy arrays instead of JAXarrays.
This should help a lot with pickling the final results.
JAX arrays can be a bit tricky to reconstruct in envs, that are not the same as the simulations were run.

So moving to numpy arrays, should enable user to pickle final results on the cluster and do the analysis (like plotting) on a local machine.

### DIFF
--- a/pysages/methods/utils.py
+++ b/pysages/methods/utils.py
@@ -87,10 +87,10 @@ class HistogramLogger:
         self.counter += 1
         if self.counter > self.offset and self.counter % self.period == 0:
             if self.data is None:
-                self.data = copy.copy(state.xi[0]).reshape((1,) + state.xi[0].shape)
+                self.data = copy.copy(state.xi[0].reshape((1,)) + state.xi[0].shape)
             else:
                 self.data = np.concatenate(
-                    (self.data, state.xi[0]).reshape((1,) + state.xi[0].shape)
+                    (self.data, state.xi[0].reshape((1,)) + state.xi[0].shape)
                 )
 
     def get_histograms(self, **kwargs):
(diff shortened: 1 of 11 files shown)


pysages/methods/utils.py
```diff
@@ -174,3 +183,35 @@ def listify(arg, replicas, name, dtype):
...
+
+    new_dict = {}
+    for key in dictionary:
+        if not numpy_only:
+            new_dict[key] = dictionary[key]
+        if isinstance(dictionary[key], str):
+            numpy_array = numpy.asarray(dictionary[key])
+            if numpy_array.dtype != numpy.dtype("O"):
+                new_dict[key] = numpy_array
+    return new_dict
```
pabloferz, line 217: I would write it like this:
```python
from pysages.utils import JaxArray

def isarray(val):
    return isinstance(val, JaxArray) or isinstance(val, numpy.ndarray)

def numpyfy_vals(dictionary: dict, arrays_only: bool = False):
    new_dict = copy.copy(dictionary) if arrays_only else {}
    for key, val in dictionary.items():
        if isarray(val):
            new_dict[key] = numpy.asarray(val)
```

InnocentBug: done

pabloferz (dismissed):

InnocentBug:
Sorry, I needed to revert your suggestion because of nested arrays.

pabloferz (approved):
It feels suboptimal trying to convert first to numpy arrays first, but let merge this for now and improve it later if possible.

PR #217, opened 2022-09-26
Add restraints in metad.py for grids

gustavor101:
In case of using grids, add restraints to metadynamics.

### DIFF
--- a/pysages/methods/metad.py
+++ b/pysages/methods/metad.py
@@ -18,6 +18,7 @@ from pysages.approxfun import compute_mesh
 from pysages.colvars import get_periods, wrap
 from pysages.grids import build_indexer
 from pysages.methods.core import GriddedSamplingMethod, Result, generalize
+from pysages.methods.restraints import apply_restraints
 from pysages.methods.utils import numpyfy_vals
 from pysages.utils import JaxArray, dispatch, gaussian, identity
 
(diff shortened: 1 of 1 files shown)


pysages/methods/metad.py
```diff
@@ -281,14 +286,23 @@ def build_bias_grad_evaluator(method: Metadynamics):
...
+            if restraints is None:
+                print("No Applying restraints")
+                force = np.zeros(grid.shape.size)
+            else:
+                print("Applying restraints")
+                lo, hi, kl, kh = restraints
+                xi, *_ = pstate
+                xi = pstate.xi.reshape(grid.shape.size)
+                force = apply_restraints(lo, hi, kl, kh, xi)
+            return force
```
pabloferz, line 305:
```suggestion
        if restraints:
            def ob_force(pstate):    # out-of-bounds force
                lo, hi, kl, kh = restraints
                xi, *_ = pstate
                xi = pstate.xi.reshape(grid.shape.size)
                force = apply_restraints(lo, hi, kl, kh, xi)
                return force
        else:
            def  ob_force(pstate):    # out-of-bounds force
                return np.zeros(grid.shape.size)
```

pysages/methods/metad.py
```diff
@@ -297,6 +311,8 @@ def evaluate_bias_grad(pstate):
             ob = np.any(np.array(pstate.grid_idx) == grid.shape)  # out of bounds
             return cond(ob, zero_force, get_force, pstate)
```
pabloferz, line 312:
```suggestion
            return cond(ob, ob_force, get_force, pstate)
```

PR #218, opened 2022-09-26
CFF free energy results shifted to zero as ABF and FUNN

gustavor101:
The cff analyze function should have the same output type (fes.max()-fes) as ABF and FUNN, this PR corrects this issue.

### DIFF
--- a/pysages/methods/ann.py
+++ b/pysages/methods/ann.py
@@ -287,8 +287,12 @@ def analyze(result: Result[ANN]):
     _, layout = unpack(model.parameters)
 
     def build_fes_fn(nn):
-        params = pack(nn.params, layout)
-        return jit(lambda x: nn.std * model.apply(params, x))
+        def fes_fn(x):
+            params = pack(nn.params, layout)
+            A = nn.std * model.apply(params, x) + nn.mean
+            return A.max() - A
+
+        return jit(fes_fn)
 
     def first_or_all(seq):
         return seq[0] if len(seq) == 1 else seq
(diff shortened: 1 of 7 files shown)


pysages/methods/ann.py
```diff
@@ -286,8 +286,12 @@ def analyze(result: Result[ANN]):
     _, layout = unpack(model.parameters)
 
     def build_fes_fn(nn):
-        params = pack(nn.params, layout)
-        return jit(lambda x: nn.std * model.apply(params, x))
+        def fes_fn(x):
+            params = pack(nn.params, layout)
+            A = nn.std * model.apply(params, x) + nn.mean
+            return A.max() - A
```
InnocentBug, line 292: Hmm, shouldn't this be `A-A.min()` to have the lowest point to be zero?

Alternative, what I also can see as reasonable is `A - A[0]` to ensure that it always starts at 0.
That makes sense for integrated free-energies for example, where you say the integration constant is 0.

pabloferz: The network is trained against the negative of the free energy, so `A.max() - A` is correct here. For now I would keep everything as we have and decide in favor another option in the future.

pysages/methods/ann.py
```diff
@@ -200,7 +200,7 @@ def learn_free_energy(state):
         nn = train(state.nn, phi)
         params = pack(nn.params, layout)
         phi = nn.std * model.apply(params, inputs).reshape(phi.shape)
-        phi = phi - phi.min()
+        phi = phi.max() - phi
```
pabloferz, line 203: I think this needs should be kept as it was. The only change needed should be the one below.

pabloferz (approved):

Issue #225, opened 2022-10-14
Repeated histogram counts if add a openmm log reporter


yihengwuKP:
# Brief Description
I found that if I add a log reporter to unbiased alanine-dipeptide simulation, the histogram logger would record one time more xi values than the time steps. 

# Environment
I followed the set up in [installation note](https://hackmd.io/Jbpc1E2kRbKPLUnmLKwINA#Installation-of-PySAGES-with-OpenMM-as-engine-on-midway2) in midway3. I ran my job in sinteractive mode in `gpu` partition.

# Code to reproduce
(this code is adapted from [example for spectral abf](https://github.com/SSAGESLabs/PySAGES/blob/main/examples/openmm/spectral_abf/alanine-dipeptide.py))
```python
#!/usr/bin/env python3

"""
SpectralABF simulation of Alanine Dipeptide in vacuum with OpenMM and PySAGES.
Example command to run the simulation `python3 alanine-dipeptide.py --time-steps 1000`
For other supported commandline parameters, check `python3 alanine-dipeptide.py --help`
"""


# %%
import argparse
import sys
import time

import matplotlib.pyplot as plt
import numpy

import pysages
from pysages.approxfun import compute_mesh
from pysages.colvars import DihedralAngle
from pysages.methods import Unbiased, HistogramLogger
from pysages.utils import try_import

openmm = try_import("openmm", "simtk.openmm")
unit = try_import("openmm.unit", "simtk.unit")
app = try_import("openmm.app", "simtk.openmm.app")


# %%
pi = numpy.pi
kB = unit.BOLTZMANN_CONSTANT_kB * unit.AVOGADRO_CONSTANT_NA
kB = kB.value_in_unit(unit.kilojoules_per_mole / unit.kelvin)

T = 298.15 * unit.kelvin
dt = 2.0 * unit.femtoseconds
adp_pdb = "adp-vacuum.pdb"


# %%
def generate_simulation(pdb_filename=adp_pdb, T=T, dt=dt):
    pdb = app.PDBFile(pdb_filename)

    ff = app.ForceField("amber99sb.xml")
    cutoff_distance = 1.0 * unit.nanometer
    topology = pdb.topology

    system = ff.createSystem(
        topology, constraints=app.HBonds, nonbondedMethod=app.PME, nonbondedCutoff=cutoff_distance
    )

    # Set dispersion correction use.
    forces = {}
    for i in range(system.getNumForces()):
        force = system.getForce(i)
[...]

Issue #226, opened 2022-10-14
Sphinix doc build failed: `[Makefile:20: html] Error 1`


yihengwuKP:
# Description
I found that using current git repo main branch, the sphinx doc build would throw this error: `[Makefile:20: html] Error 1`

# Reproduce the bug
I followed the set up in [installation note](https://hackmd.io/Jbpc1E2kRbKPLUnmLKwINA#Installation-of-PySAGES-with-OpenMM-as-engine-on-midway2) in midway3. I ran the doc build in the login node.
Also, I followed the [README.md](https://github.com/SSAGESLabs/PySAGES/blob/main/docs/README.md) in docs: 
```bash
pip install -r docs/requirements.txt
make html 
```
And these lines give a stdout like this:
```
sphinx-build -M html "source" "build" -W --keep-going -b spelling &&\                                                                                                                                                                                                                    
sphinx-build -M html "source" "build" -W                                                                                                                                                                                                                                                 
Running Sphinx v5.2.3                                                                                                                                                                                                                                                                    
Initializing Spelling Checker 7.6.1                                                                                                                                                                                                                                                      
loading pickled environment... done                                                                                                                                                                                                                                                      
Ignoring wiki words                                                                                                                                                                                                                                                                      
Ignoring acronyms                                                                                                                                                                                                                                                                        
Ignoring Python builtins                                                                                                                                                                                                                                                                 
Ignoring importable module names                                                                                                                                                                                                                                                         
Ignoring contributor names                                                                                                                                                                                                                                                               
Looking for custom word list in /home/yihengwu917/source_code/PySAGES/docs/source/pysages_wordlist.txt                                                                                                                                                                                   
Scanning contributors                                                                                                                                                                                                                                                                    
[autosummary] generating autosummary for: getting-started.rst, index.rst, installation.rst, module-pysages-colvars-angles.rst, module-pysages-colvars-coordinates.rst, module-pysages-colvars-core.rst, module-pysages-colvars-shape.rst, module-pysages-colvars.rst, module-pysages-meth
ods-abf.rst, module-pysages-methods-ann.rst, ..., module-pysages-methods-funn.rst, module-pysages-methods-harmonic_bias.rst, module-pysages-methods-metad.rst, module-pysages-methods-string.rst, module-pysages-methods-umbrella.rst, module-pysages-methods-utils.rst, module-pysages-m
[...]

PR #227, opened 2022-10-14
Phase angle

yihengwuKP:
# Brief Description
Implementation of cv: phase angle of monocyclic ring molecules, using [Cremer-Pople method](https://pubs.acs.org/doi/10.1021/ja00839a011). Notice that for ring with more than 5 members, there are (N-1)/2-1 phase angles defined in Cremer-Pople method, and this class only calculate the first one (m=2 in Eq 12 and Eq 13). Potentially m can be added as an argument but I haven't figured out how, should I add an `__init__` for the class `PhaseAngle`? I would appreciate some help on that! However, this is already enough for my usage. 

# Test
I've tested this implementation against the implementation in [`MDAnalysis.nuclinfo.phase_cp`](https://docs.mdanalysis.org/1.0.1/documentation_pages/analysis/nuclinfo.html). I found that it's tricky actually to install MDAnalysis into the pysages conda environment in midway3, as the MDAnalysis would install one version of scipy that messed up with jax. So I used another conda environment for the analysis.

## Code for testing
```python
#!/usr/bin/env python

import openmm as mm
import openmm.app as app
from openmm import unit
from math import ceil
import os
import numpy as np
from pysages.colvars.angles import PhaseAngle, DihedralAngle
from pysages.methods import HistogramLogger, Unbiased
import pysages

step_size = 1 * unit.femtosecond
nsteps = 1000
print("nsteps = {}".format(nsteps))

pdb = app.PDBFile('guanosine.fixed.pdb')
sugar_indices = {}
for i, atom in enumerate(pdb.topology.atoms()):
    if atom.name in ["O4'", "C4'", "C3'", "C2'", "C1'"]:
        sugar_indices[atom.name] = atom.index

print(f"dictionary of sugar indices: {sugar_indices}")


def generate_simulation():

    forcefield = app.ForceField('amber14-all.xml', 'amber14/tip3pfb.xml')
    system = forcefield.createSystem(pdb.topology, nonbondedMethod=app.PME,
                                     nonbondedCutoff=1.2*unit.nanometer,
                                     constraints=app.HBonds)
    system.addForce(mm.MonteCarloBarostat(1*unit.bar,
                                          298*unit.kelvin,
                                          1000))

    integrator = mm.LangevinIntegrator(298*unit.kelvin, 5/unit.picosecond,
                                       step_size.in_units_of(unit.picosecond))
    simulation = app.Simulation(pdb.topology, system, integrator)
    simulation.context.setPositions(pdb.positions)

    print("minimizing energy...")
    simulation.minimizeEnergy()

    print("Using {} platform".format(simulation.context.getPlatform().getName()))
    simulation.reporters.append(app.DCDReporter('output.dcd', 1))

    return simulation


cvs = [PhaseAngle(
        [sugar_indices[nm] for nm in ["O4'", "C1'", "C2'", "C3'", "C4'"]])]
# notice that the order of the indices do matter for the calculation of phase angle
method = Unbiased(cvs)
callback = HistogramLogger(1)

raw_result = pysages.run(method, generate_simulation, nsteps, callback)
np.savetxt('phase_angle.txt', raw_result.callbacks[0].data)
```
the input and output files and the scripts: [phase_angle_example.zip](https://github.com/SSAGESLabs/PySAGES/files/9788453/phase_angle_example.zip)

Comparing the results:
```python

import numpy as np
import MDAnalysis as mda
import matplotlib.pyplot as plt
from MDAnalysis.analysis import nuclinfo
from math import sin, cos, atan2, pi, sqrt

myP = np.loadtxt("phase_angle.txt")

u = mda.Universe("guanosine.fixed.pdb", "output.dcd")
P = []
for t in u.trajectory:
    P.append(nuclinfo.phase_cp(u, 'A', 1))

myP=(myP+np.pi/2)*180/np.pi%360
plt.plot(myP[1:], 'o', label='pysages', alpha=.5)
plt.plot(np.array(P[:-1]), label='MDAnalysis')
plt.legend()
plt.show()


plt.figure(figsize=(5,5))
plt.scatter(myP[1:], P[:-1])
plt.plot([320, 350], [320, 350], color='red')
```
the result: 
image
image

There is this manual hack of myP[1:], P[:-1], because I found the logger result from histogromLogger seems to be shifted, related to issue #225


yihengwuKP:
I think this failed CI/doc-test is related with issue #226

yihengwuKP:
Ok, I've added another commit to fix the spelling issue!

docs/source/pysages_wordlist.txt
```diff
@@ -48,3 +48,9 @@ SimulationContext
 decompositions
 conda
 numpyfying
+monocyclic
+Cremer
+five-membered
```
pabloferz, line 53:
```suggestion
membered
```

pysages/colvars/angles.py
```diff
@@ -121,3 +122,80 @@ def dihedral_angle(p1, p2, p3, p4):
...
+
+
+class PhaseAngle(CollectiveVariable):
+    """
+    Computes the phase angle of a monocyclic ring by the Cremer-Pople method.
+    Mathematical definitions can be found in
+    [D. Cremer and J. A. Pople, JACS, 1974]
+    (https://pubs.acs.org/doi/10.1021/ja00839a011)
+    Equation 4-14
+    Notice that for ring with more than 5 members, there are (N-1)/2-1 phase angles, and
```
ricalessandri, line 134:
```suggestion
    Notice that for ring with more than 5 atoms, there are (N-1)/2-1 phase angles, and
```

pysages/colvars/angles.py
```diff
@@ -121,3 +122,80 @@ def dihedral_angle(p1, p2, p3, p4):
...
+    this class only calculate the first one (m=2 in Equation 12 and Equation 13, or see
```
ricalessandri, line 135:
```suggestion
    this class only calculates the first one (m=2 in Equation 12 and Equation 13, or see
```

pysages/colvars/angles.py
```diff
@@ -121,3 +122,80 @@ def dihedral_angle(p1, p2, p3, p4):
...
+    `pysages.colvars.angles.phase_angle` for math).
+    Additional details and implementation can be found in
+    [MDAnalysis.analysis.nuclinfo.phase_cp]
+    (https://docs.mdanalysis.org/1.0.1/documentation_pages/analysis/nuclinfo.html),
+    where it specifically calculates the phase angle for ribose, a five-membered ring.
+    (Also, notice that in this implementation of MDAnalysis, a 90 degree is added to the resulted
```
ricalessandri, line 141:
```suggestion
    (Also, notice that in this implementation of MDAnalysis, a 90 degree is added to the resulting
```

pysages/colvars/angles.py
```diff
@@ -121,3 +122,80 @@ def dihedral_angle(p1, p2, p3, p4):
...
+    phase angle by CP method, because it's convention to add 90 degree to convert to
```
ricalessandri, line 142:
```suggestion
    phase angle by CP method, because it is convention to add 90 degrees to convert to the
```

pabloferz:
Yes, you should be able to commit the suggestions, but you can also manually make the changes on your working branch and push them here.

pysages/colvars/angles.py
yihengwuKP: ah, this makes sense to me!

pysages/colvars/angles.py
```diff
@@ -121,3 +122,76 @@ def dihedral_angle(p1, p2, p3, p4):
...
+    """
+    Computes the phase angle of a monocyclic ring by the Cremer-Pople method.
+    Mathematical definitions can be found in
+    [D. Cremer and J. A. Pople, JACS, 1974]
+    (https://pubs.acs.org/doi/10.1021/ja00839a011)
+    Equation 4-14
+    Notice that for rings with more than 5 atoms, there are (N-1)/2-1 phase angles, and
+    this class only calculates the first one (m=2 in Equation 12 and Equation 13, or see
+    `pysages.colvars.angles.phase_angle` for math).
+    Also, the phase angle obtained from Cremer-Pople method can be convert to
```
ricalessandri, line 137:
```suggestion
    Also, the phase angle obtained via the Cremer-Pople method can be converted to the
```

pysages/colvars/angles.py
```diff
@@ -121,3 +122,76 @@ def dihedral_angle(p1, p2, p3, p4):
...
+    Mathematical definitions can be found in
+    [D. Cremer and J. A. Pople, JACS, 1974]
+    (https://pubs.acs.org/doi/10.1021/ja00839a011)
+    Equation 4-14
+    Notice that for rings with more than 5 atoms, there are (N-1)/2-1 phase angles, and
+    this class only calculates the first one (m=2 in Equation 12 and Equation 13, or see
+    `pysages.colvars.angles.phase_angle` for math).
+    Also, the phase angle obtained via the Cremer-Pople method can be converted to the
+    Altona-Sundaralingam definition by adding pi/2 to the result and then
+    convert to 0 to 360 range.
```
pabloferz, line 139:
```suggestion
    [D. Cremer and J. A. Pople, JACS, 1974](https://pubs.acs.org/doi/10.1021/ja00839a011)
    Equations 4-14.
    Notice that for rings with more than 5 atoms, there are `(N - 1) / 2 - 1`
    phase angles, and this class only calculates the first one (`m = 2` in
    Equations 12 and 13, or see `pysages.colvars.angles.phase_angle` for math).
    Also, the phase angle obtained via the Cremer-Pople method can be converted
    to the Altona-Sundaralingam order parameter by adding `pi / 2` to the result
    and then converting to degrees.
```

pysages/colvars/angles.py
```diff
@@ -121,3 +122,76 @@ def dihedral_angle(p1, p2, p3, p4):
...
+    r"""
+    calculate phase angle (first phase angle if N>5) based on Cremer-Pople method.
+    :math:`r0 = 1/N \sum_i^N \vec{r}_i`
+    :math:`\vec{R1} = \sum_i^N (\vec{r}_i -r_c) \sin(2\pi (i-1)/N)`
+    :math:`\vec{R2} = \sum_i^N (\vec{r}_i -r_c) \cos(2\pi (i-1)/N)`
+    :math:`\hat{n} = \vec{R1} \times \vec{R2}/ (|\vec{R1}\times\vec{R2}|)`
+    :math:`z_i = (\vec_{r}_i-r_c) \cdot \hat{n}`
+    :math:`a =  \sqrt(2/N) \sum_i^N z_i \cos(2\pi 2(i-1)/N)`
+    :math:`b = -\sqrt(2/N) \sum_i^N z_i \sin(2\pi 2(i-1)/N)`
+    :math:`P=\atan2(b/a)`
```
pabloferz, line 170:
```suggestion
    :math:`P = \atan(b / a)`
```

pabloferz (dismissed):

yihengwuKP:
Oh, it seems I somehow dismissed Pablo's stale review? So I shouldn't have pushed another commit? I just implement another relavant cv, that's the square root of the a and b which are calculated during the calculation of the phase angle. Not sure if this implementation is the most efficient way of doing it though. This c.v. is the amplitude of the ring pucker and less useful than the phase angle, but might be good to have.

InnocentBug:
No, this is not an issue. 
you automatically dismiss the review when you push changes. 
It just means that we have to review the changed version before we can merge. 

It doesn’t mean that you disregard his recommendations. More the opposite. 
when you are done with changes, just rerequest the review.

yihengwuKP:
Thanks! I've added the doc string and it's now ready for review

pabloferz:
After looking at the two CVs, I think there's a lot of room for simplifying the implementation. We should probably provide three classes here:

 1. A class called `RingPuckeringCoordinates`, this would be a multicomponent CV (see https://github.com/SSAGESLabs/PySAGES/blob/main/pysages/colvars/coordinates.py#L113)
 2. A class `RingPhaseAngle` whose `function` should call `ring_pluckering_coordinates`,
 3. And a class `RingAmplitude`, again, with `ring_amplitude` calling `ring_pluckering_coordinates`.

This should good enough for now even if we do not provide yet a way to compute the coordinates for each value of `m` in the Cremer-Pople method.

yihengwuKP:
Ah, thanks for the guidance! This is what I don't know, and I'll try to implement it.

pysages/colvars/angles.py
```diff
@@ -121,3 +122,142 @@ def dihedral_angle(p1, p2, p3, p4):
...
+
+
+class PhaseAngle(CollectiveVariable):
+    """
+    Computes the phase angle of a monocyclic ring by the Cremer-Pople method.
+    Mathematical definitions can be found in
+    [D. Cremer and J. A. Pople, JACS, 1974](https://pubs.acs.org/doi/10.1021/ja00839a011)
+    Equations 4-14.
+    Notice that for rings with N atoms, there are `int( ( N - 1 )  / 2 - 1 )` phase angles.
+    So if the ring contains more than six atoms, there are more than one phase angle.
```
ricalessandri, line 134:
```suggestion
    So if the ring contains more than six atoms, there is more than one phase angle.
```

pysages/colvars/angles.py
```diff
@@ -121,3 +122,142 @@ def dihedral_angle(p1, p2, p3, p4):
...
+
+
+class AmplitudeRing(CollectiveVariable):
+    """
+    Computes the amplitude of a monocyclic ring by the Cremer-Pople method.
+    Mathematical definitions can be found in
+    [D. Cremer and J. A. Pople, JACS, 1974](https://pubs.acs.org/doi/10.1021/ja00839a011)
+    Equations 4-14.
+    Notice that for rings with N atoms, there are `int (N / 2 - 1) ` amplitudes.
+    So if the ring contains more than five atoms, there are more than one amplitude.
```
ricalessandri, line 207:
```suggestion
    So if the ring contains more than five atoms, there is more than one amplitude.
```

pysages/colvars/angles.py
```diff
@@ -121,3 +122,142 @@ def dihedral_angle(p1, p2, p3, p4):
...
+    This class (for now) only calculates the first one (`m = 2` in
+    Equations 12 and 13, or see `pysages.colvars.angles.phase_angle` for math).
+    Also, the amplitude obtained via the Cremer-Pople method can be converted
+    to the Altona-Sundaralingam order parameter by multiplying `102.5` to the result.
```
ricalessandri, line 211:
```suggestion
    to the Altona-Sundaralingam order parameter by multiplying the result by `102.5`.
```

yihengwuKP:
Ok, I've implemented the multicomponent cv RingPuckeringCoordinates like @pabloferz suggested. I've tested them against MDAnalysis calculations and results are following:
image
image
image
image

InnocentBug:
Ok, that looks great.
Awesome that you have tests for your code.
I know that you said, you have troubles with analysis and running the code.

But can you try to sort your test code into the `examples/openmm` directory.
And after that, please add your example to the automated testing in `.github/workflows/docker-ci.yml`.
You can find many examples there that make use of example scripts.
It might be that you have to prune the example a little bit in order to complete the simulation in about 15min on a single cpu.

It is OK if that is too short for actual physical testing, but we want to run your code contribution to make sure that it works fine.

Thank you.

yihengwuKP:
cool, right now the example runs a few seconds in gpu. I'll try to include it in the example and the CI. I'll also fix the doc-test issues.

yihengwuKP:
I didn't add it to `examples` because I thought they are all for the methods, should I just add a `openmm/unbiased` example and put my scripts there?

pysages/colvars/angles.py
```diff
@@ -121,3 +127,152 @@ def dihedral_angle(p1, p2, p3, p4):
...
+    `int( ( N - 1 )  / 2 - 1 )` phase angles and `int (N / 2 - 1) ` amplitudes.
+    So if the ring contains more than six atoms, there is more than one phase angle;
+    similarly, if the ring contains more than five atoms, there is more than one amplitude.
+    This class (for now) only calculates the first amplitude and the phase angle (`m = 2` in
+    Equations 12 and 13, or see `pysages.colvars.angles.ring_puckering_coordinates` for math).
+    Also, the phase angle obtained via the Cremer-Pople method can be converted
+    to the Altona-Sundaralingam order parameter by adding `pi / 2` to the result
+    and then converting from radian to degrees.
+    Similarly, the amplitude obtained via the Cremer-Pople method can be converted
+    to the Altona-Sundaralingam order parameter by multiplying the result by `1025`.
```
ricalessandri, line 149: 1025 or 102.5?

InnocentBug:
Oh, I see.
I think the reason all examples are methods, is because that was the main focus of development.
I would include the example as `openmm/pucker`, so the CV instead of the method.

yihengwuKP:
Cool, thanks for all your help! I'll do that

pysages/colvars/angles.py
ricalessandri: ok!

pysages/colvars/angles.py
```diff
@@ -121,3 +127,152 @@ def dihedral_angle(p1, p2, p3, p4):
...
+    So if the ring contains more than six atoms, there is more than one phase angle;
+    similarly, if the ring contains more than five atoms, there is more than one amplitude.
+    This class (for now) only calculates the first amplitude and the phase angle (`m = 2` in
+    Equations 12 and 13, or see `pysages.colvars.angles.ring_puckering_coordinates` for math).
+    Also, the phase angle obtained via the Cremer-Pople method can be converted
+    to the Altona-Sundaralingam order parameter by adding `pi / 2` to the result
+    and then converting from radians to degrees.
+    Similarly, the amplitude obtained via the Cremer-Pople method can be converted
+    to the Altona-Sundaralingam order parameter by multiplying the result by `1025`.
+    (assume the unit is nanometer for the amplitude, and this will convert from nanometer to degree)
```
ricalessandri, line 150:
```suggestion
```

examples/openmm/cv_pucker/unbiased.py
```diff
@@ -0,0 +1,57 @@
...
+    return simulation
+
+
+cvs = [RingPuckeringCoordinates([sugar_indices[nm] for nm in ["O4'", "C1'", "C2'", "C3'", "C4'"]])]
+# notice that the order of the indices do matter for the calculation of phase angle
+method = Unbiased(cvs)
+callback = HistogramLogger(1)
+
+raw_result = pysages.run(method, generate_simulation, nsteps, callback)
+np.savetxt("phase_angle.txt", raw_result.callbacks[0].data)
```
ricalessandri, line 57: This part could probably go into a `def main` function, like, e.g., [here](https://github.com/SSAGESLabs/PySAGES/blob/main/examples/openmm/spectral_abf/alanine-dipeptide.py)

pysages/colvars/angles.py
```diff
@@ -121,3 +127,152 @@ def dihedral_angle(p1, p2, p3, p4):
...
+    `int( ( N - 1 )  / 2 - 1 )` phase angles and `int (N / 2 - 1) ` amplitudes.
+    So if the ring contains more than six atoms, there is more than one phase angle;
+    similarly, if the ring contains more than five atoms, there is more than one amplitude.
+    This class (for now) only calculates the first amplitude and the phase angle (`m = 2` in
+    Equations 12 and 13, or see `pysages.colvars.angles.ring_puckering_coordinates` for math).
+    Also, the phase angle obtained via the Cremer-Pople method can be converted
+    to the Altona-Sundaralingam order parameter by adding `pi / 2` to the result
+    and then converting from radians to degrees.
+    Similarly, the amplitude obtained via the Cremer-Pople method can be converted
+    to the Altona-Sundaralingam order parameter by multiplying the result by `1025`.
```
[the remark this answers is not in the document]
yihengwuKP: cool, that's much clearer!

examples/openmm/cv_pucker/unbiased.py
yihengwuKP: ah, right, totally! I was a bit lazy

yihengwuKP: I'll change it

.github/workflows/docker-ci.yml
```diff
@@ -143,6 +143,24 @@ jobs:
...
+        uses: actions/download-artifact@v2
+        with:
+          name: pysages
+          path: /tmp
+      - name: Load and run test
+        run: |
+          docker load --input /tmp/pysages.tar
+          SCRIPT="cd PySAGES/examples/openmm/cv_pucker &&"
+          SCRIPT="${SCRIPT} python unbiased.py" 
+          docker run -t pysages bash -c "cd PySAGES/examples/openmm/cv_pucker && python3 ./unbiased.py"
```
InnocentBug, line 162: Is there a way to pre-record the expected CV and compare with the PySAGES result.

yihengwuKP: It's hard to make it purely deterministic. However, I have scripts to do post-hoc calculation using MDAnalysis and can compare. I guess I can change the unbiased.py, like adding an assert on the deviation between two results?

InnocentBug: Yeah if you can do that, that works too.
Isn't there a way to set all seeds and then it is determinstic?
Computers are deterministic machines after all.

Whatever works.

InnocentBug: https://github.com/SSAGESLabs/PySAGES/actions/runs/3298840961/jobs/5441644009

yihengwuKP: cool I'll do that. I said that because I have a vague memory telling me that openmm might not be totally trustworthy on that. I found similar quote in their documentation: 

> 10.4. Determinism
> Whether a simulation is deterministic will depend on what plaform you run on in addition to what settings/methods you use. For instance, as of this writing, using PME on the Reference, OpenCL, and double-precision CUDA will result in deterministic simulations. Single-precision CUDA and CPU platforms are not deterministic in this case. However, none of this behavior is guaranteed in future versions. In many cases it will still result in an identical trajectory. If determinism is a critical for your needs, you should carefully check to ensure that your settings and platform allow for this.

I am just lazy to deal with that lol. But anyways, I'll add it to the script!

InnocentBug: sure, I understand, anyway of checking it is doing the right thing is fine

yihengwuKP:
ah, it seems that I can't put my input PDB into `examples/inputs` because now the gitignore ignores it `example/**/*.pdb`. Any suggestions on quickly fixing this without hacking?

yihengwuKP:
Ah, I see, I'll use `git add -f`

pysages/colvars/angles.py
```diff
@@ -121,3 +127,152 @@ def dihedral_angle(p1, p2, p3, p4):
...
+    R2 = np.dot(rc.T, np.real(fourier_coeff))
+    n = np.cross(R1, R2)
+    n /= linalg.norm(n)
+    z = np.dot(rc, n)
+    fourier_coeff2 = np.exp(2 * theta)
+    a = np.sqrt(2 / N) * np.sum(z * np.real(fourier_coeff2))
+    b = -np.sqrt(2 / N) * np.sum(z * np.imag(fourier_coeff2))
+    q = np.sqrt(a**2 + b**2)
+    phi = np.arctan2(b, a)
+    return np.array([q, phi])
```
pabloferz, line 206: What I had in mind was to return `a` and `b` here and compute `q` and `phi` on the `RingAmplitude` and `RingPhaseAngle`, respectively:
```suggestion
    return np.array([a, b])
```

pysages/colvars/angles.py
```diff
@@ -121,3 +127,152 @@ def dihedral_angle(p1, p2, p3, p4):
...
+    rs: DeviceArray
+        :math: `\vec{r}_i` array of 3D vector in space
+
+    Returns
+    ------------
+    float
+        :math:`P` in range -pi to pi.
+    """
+    _, phi = ring_puckering_coordinates(rs)
+    return phi
```
pabloferz, line 242:
```suggestion
    a, b = ring_puckering_coordinates(rs)
    return np.arctan2(b, a)
```

yihengwuKP: I see, but there might be a need for people to get both the q and the phi?

pabloferz: That's fair. Not sure what would the most common scenario would be. Do you have any opinions on this @ricalessandri.

yihengwuKP: I think the most useful one is the phase angle with the real info, the amplitude, at least in the 5-membered ring in RNA, is usually calculated with the phase angle. I haven't seen people who only calculate the amplitude, because usually, it fluctuates around some eq value, but phase angle can switch between different conformations.

pabloferz: Ok. Let's leave as you have it now, and we will revisit if/when there's a demand for something more flexible.

pabloferz (approved):

pysages/colvars/angles.py
yihengwuKP: cool, thanks for the review! I'll try to add the test that Ludwig suggested today!

yihengwuKP:
OK, I've added the testing for phase angle like @InnocentBug had suggested.

examples/openmm/cv_pucker/phase_angle.txt
```diff
@@ -0,0 +1,100 @@
+2.635088190436363220e-02 -1.986930370330810547e+00
```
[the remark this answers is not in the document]
InnocentBug: Same with output.dcd, that shouldn't be tracked either

examples/openmm/cv_pucker/sugar_coords.txt
```diff
@@ -0,0 +1,100 @@
+3.365954756736755371e-01 -8.431397080421447754e-01 1.030287742614746094e+00 2.687700688838958740e-01 -7.988463044166564941e-01 9.136055111885070801e-01 3.582888841629028320e-01 -8.384060859680175781e-01 7.957995533943176270e-01 5.004750490188598633e-01 -8.411503434181213379e-01 8.533967733383178711e-01 4.749850630760192871e-01 -8.646225929260253906e-01 1.003347277641296387e+00
```
InnocentBug, line 1: this is also an output and should be tracked.

.github/workflows/docker-ci.yml
InnocentBug, line 162: you do the work twice here.
you define the SCRIPT variable, but then you also write down the command again.
One is fine.

InnocentBug (changes requested):

InnocentBug:
Thanks for adding the test.
But we need to make sure to not track any output in git repos.
And so far, I don't think you are actually executing the test after the simulation run is done.

examples/openmm/cv_pucker/phase_angle.txt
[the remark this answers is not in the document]
yihengwuKP: ah, good point

.github/workflows/docker-ci.yml
yihengwuKP: thanks for the code review, I'll fix it

yihengwuKP:
That's true, I tested it locally and got excited and forgot about it, I'll fix everything soon.

InnocentBug (approved):

pabloferz:
Thanks for this @yihengwuKP!

PR #231, opened 2022-10-27
Add GeM CV

maggiezimon:
Implementation of the Geometry Matching (GeM) Local Order Metric (LOM). The algorithm enabling the measurement of order in the neighborhood of an atomic or molecular site is described in [Martelli2018](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.97.064105).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/mwouts/jupytext
-    rev: v1.14.1
+    rev: v1.14.2
     hooks:
       - id: jupytext
         args: [--sync]
--- a/examples/Advanced_Sampling_Introduction.md
+++ b/examples/Advanced_Sampling_Introduction.md
@@ -7,7 +7,7 @@ jupyter:
       extension: .md
       format_name: markdown
       format_version: '1.3'
-      jupytext_version: 1.14.1
+      jupytext_version: 1.14.2
   kernelspec:
     display_name: Python 3
     name: python3
--- a/examples/Install_PySAGES_Environment.md
+++ b/examples/Install_PySAGES_Environment.md
@@ -7,7 +7,7 @@ jupyter:
       extension: .md
       format_name: markdown
       format_version: '1.3'
-      jupytext_version: 1.14.1
+      jupytext_version: 1.14.2
   kernelspec:
     display_name: Python 3
     name: python3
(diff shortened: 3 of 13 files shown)


maggiezimon:
The checks have failed because we are loading `jax_md` module.

pabloferz:
We test that our methods and collective variables are pickleable here https://github.com/SSAGESLabs/PySAGES/blob/main/tests/test_pickle.py. We would need an example call to the constructor of GeM to make sure that this is true for it as well (or find out how to make `GeM` pickleable if not).

maggiezimon:
@pabloferz It is not very pickable due to the dependency on jax_md. In `test_pickle.py` we should have something like this

```
snapshot_positions = np.array([
[2.0e-03, -2.40e-02, 5.0e-03],
[3.2e-01, 3.23e-01, 1.7e-02],
[3.14e-01,-2.10e-02, 3.16e-01],
[3.0e-03, 3.16e-01, 3.02e-01],
[1.58e-01, 1.37e-01, 1.62e-01],
[4.74e-01, 5.02e-01, 1.65e-01],
[4.84e-01, 1.58e-01, 4.54e-01],
[1.65e-01, 4.57e-01, 4.81e-01]])

COLVAR_ARGS = {
    "GeM": {"indices": np.arange(10), 
                   "reference_positions": np.array(
                   [[1.0, 1.0, 1.0], 
                   [-1.0, -1.0, 1.0],
                   [-1.0, 1.0, -1.0],
                   [1.0, -1.0, -1.0]],
                   "box": np.array(
                   [[2.0, 0., 0.],
                   [0., 2.0, 0.],
                   [0., 0., 2.0]]), 
                   "number_of_rotations": 20, "number_of_opt_it": 10, 
                   "standard_deviation": 0.125 "mesh_size": 30, 
                   "nbrs": gen_neighbor_list(2.0, positions=snapshot_positions, r_cutoff=1.5)), 
                   "fractional_coords": True},
}
```
The problem is `nbrs` which is a callable JAX MD neighbor list function to update the neighbor list. It could be implemented as follows:
```
from jax_md.partition import neighbor_list as nlist, NeighborListFormat, space

def gen_neighbor_list(box_size, positions, r_cutoff, cap_mult):
    '''Generate a jax-md function
    for updating the neighbor list'''
    displacement_fn, shift_fn = space.periodic(box_size)
    neighbor_list_fn = nlist(
        displacement_fn, box_size, r_cutoff, capacity_multiplier=cap_mult,
        format=NeighborListFormat.Dense)
    neighbors = neighbor_list_fn.allocate(positions)
    return neighbors
```
However, this way, `nbrs` is a function of positions and some parameters. Please, let me know how you would like me to change it.

PR #243, opened 2023-01-18
Move linters config files to .trunk/config

pabloferz:
I'd like to have less clutter on the file structure shown in GH, and this is supported by trunk.


InnocentBug (changes requested):
Generally, good plan! It's a feature of trunk that they added recently.

However, I think the `.readthedocs.yaml` needs to stay where it was.
That is a file for readthedocs, not trunk or linting.

InnocentBug:
Please don't force push, after I start reviewing.
Makes it very hard for me to keep track of what you changed after my comment.
Thanks

pabloferz:
I did force push just because of the way my local configuration is setup, but did not modify the commit structure, you should be able to see the new changes without any issue.

InnocentBug (approved):

PR #245, opened 2023-01-19
Update copyright year and tweak file headers

pabloferz:
This should make simpler to update the copyright notice by only modifying the `LICENSE.md` file.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,14 +1,14 @@
 # License
 
-PySAGES as a whole is licensed under the GPL-3.0 or later. But do note that only some of
-the code under the `methods` is covered by GPL, as it is based on the original SSAGES
-implementation. The rest of the code is licensed under the MIT License.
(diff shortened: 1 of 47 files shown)


InnocentBug (approved):
I am not entirely sure about the legal implications of removing the copyright line from the individual files.
But if that is OK, this looks good.

pabloferz:
All files are covered by the LICENSE as being part of the project even if the have no explicit notice.

The practice of putting a notice in every file is usually done in case the individual files end up being shared. A very light way around this issue is to use [SPDX identifiers](https://spdx.dev/ids/) as we do (and other projects such as the [Linux Kernel](https://www.kernel.org/doc/html/latest/process/license-rules.html#license-identifier-syntax)). What we are doing is even better than just the SPDX IDs since we are pointing explicitly to where people can find the license, and the license contains the copyright notice.

pabloferz:
The trunk failure was an issue on GH and unrelated to the changes.

PR #249, opened 2023-01-27
Fix analyze for ABF and FUNN

pabloferz:
I had missed removing the mean from the data before training the NN in both `analyze` implementations.

### DIFF
--- a/pysages/methods/abf.py
+++ b/pysages/methods/abf.py
@@ -12,29 +12,22 @@ which can be integrated to yield the free energy.
 
 The implementation of the adaptive biasing force method here closely follows
 https://doi.org/10.1063/1.2829861. One important difference is that the time
-derivative of the product :math:`W\\cdot p` (equation 9 of reference) is approximated by a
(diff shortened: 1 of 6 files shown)


pabloferz:
I have validated this with butane and ADP and seems to work fine. If someone has a saved result and a working environment, this should be easy to test.

ricalessandri (approved):
I've tested this branch on alanine dipeptide pickled data that I had from previous runs (which btw I could load and re-analyze simply with:
```
raw_result = pickle.load( open("raw_result.pickle", "rb") )
result = pysages.analyze(raw_result)
```
).

The implementation seems indeed to work fine. See this comparison of a previous FUNN run, namely:

1. (left) previous analyze version
2. (center) re-analyze data with this `pz/analyze` branch
3. (right) free energy obtained with the SSAGES `ABF_integrator.py`

Screen Shot 2023-02-14 at 3 59 53 PM

I think the agreement between center and right is practically perfect given the different methods used to integrate.

PR #250, opened 2023-01-31
FFS Coarse Graining option

gustavor101:
Add Coarse Graining FFS (https://doi.org/10.1073/pnas.1509267112) with the option of different timesteps for initial flow and the window sampling functions.


InnocentBug:
Please run `./trunk fmt` on this branch to make the error go away.

PR #254, opened 2023-02-28
Interface for restarts

pabloferz:
This adds two ways of restarting runs:

 1. From a stored `Result` (this is compatible with our MPI interface). One change introduced here is that the last snapshot of a simulation is now also part of `Result` instances. The interface looks like
    `pysages.run(result: Result, context_generator: Callable, timesteps: int, *other_args, **kwargs)`
    (that is, one would switch `method` by `result` when restarting).

 2. From a `SamplingContext` (which is a renaming of our `ContextWrapper`). This is useful for interactive sessions, say in a jupyter notebook, or for debugging purposes.

FFS is left out for now as it is too different and requires other changes to its interface.


gustavor101 (changes requested):
The restart functionality works well with pickle files and multiple calls in notebooks, but a couple of examples could be useful for the user like showing that the loggers are  now incorporated in SamplingContext and not in pysages.run as before.

InnocentBug (dismissed):

gustavor101 (approved):

PR #256, opened 2023-03-01
More fixes for the Google Collab notebooks

InnocentBug:
Curiously, there is a problem with the `plum.dispatch` of constructors.
If I specify arguments with keywords, a constructor is called that doesn't match any on the ones specified and I get basically an empty `UmbrellaIntegration` instance.
That is weird, and potentially problematic. But I don't know how to fix it.

Not using keywords in this case fixed the problem.

### DIFF
--- a/examples/Advanced_Sampling_Introduction.ipynb
+++ b/examples/Advanced_Sampling_Introduction.ipynb
@@ -151,9 +151,7 @@
     "\n",
     "rm -rf PySAGES\n",
     "git clone https://github.com/SSAGESLabs/PySAGES.git &> /dev/null\n",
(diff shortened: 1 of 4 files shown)


InnocentBug:
https://github.com/beartype/plum/issues/40

This is underlying issue behind what we see here.

pabloferz (approved):

pabloferz:
The issue now is the newest version of `plum`. I need to fix that separately. This looks fine to me so I will go ahead and merge.

pabloferz:
Just to let some breadcrumbs here. Everything should work fine by installing a plum version <2. E.g., `pip install "plum-dispatch<2"`

Issue #258, opened 2023-03-17
Error in the example of alanine dipeptide using colab


Chenghao-Wu:
I am trying to test the example of alanine dipeptide in colab but met the following error:

it happens when I tried to import pysages:

from pysages.grids import Grid
from pysages.colvars import DihedralAngle
from pysages.methods import Metadynamics

**import pysages**

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
[<ipython-input-13-93d8866a310a>](https://localhost:8080/#) in <module>
      2 from pysages.colvars import DihedralAngle
      3 from pysages.methods import Metadynamics
----> 4 import pysages

[/usr/local/lib/python3.9/dist-packages/pysages/__init__.py](https://localhost:8080/#) in <module>
     72 from .utils import dispatch  # noqa: E402, F401
     73 
---> 74 run = dispatch._functions["run"]
     75 analyze = dispatch._functions["analyze"]
     76 

AttributeError: 'Dispatcher' object has no attribute '_functions'

Thank you very much if any suggestion can be provided!

pabloferz:
One of our dependencies got updates that break pysages. We are working on a more permanent fix, but for now you can add a cell (after the one where pysages is installed) with the following:

```
!pip install "plum-dispatch<2"
```

Chenghao-Wu:
Thank you very much for the update. It works well this time!

PR #257, opened 2023-03-17
Fix compatibility with plum>=2

pabloferz:
`plum >=2` brings some breaking changes which cause tests to fail right now. This PR provides a compatibility layer to support the newest plum version while preserving compatibility with `plum <2`.

/trunk merge

PR #259, opened 2023-03-18
Restrict plum to <2

pabloferz:
Stopgap solution before #257 can be merged. This should allow people to use pysages while we sort out the issues with supporting plum 2.


pabloferz:
Fixes #258.

PR #260, opened 2023-03-20
Make Sampler a subclass of ASE's Calculator

pabloferz:
The way ASE integrates a MD simulation heavily relies on that any force applied should come from a `ase.calculator.Calculator`. Making our `pysages.backends.ase.Sample` inherit from the former corrects an overestimate on the free energy we have been observing for ASE runs.

CC @xuyinan2


gustavor101 (approved):
it is working well.

PR #277, opened 2023-08-13
Fix a couple of lammps backend bugs

pabloferz:


ndtrung81 (approved):
thanks @pabloferz How did you get the bug with the masses, that needs to be reshape here?

pabloferz:
>thanks @pabloferz How did you get the bug with the masses, that needs to be reshape here?
I tried running an ABF simulation, which requires the momenta (this was not the case with the unbiased simulations), but we need the same dimensionality on both the velocities and masses so each particle velocity is multiplied by the corresponding mass.

Issue #278, opened 2023-09-04
`pysages.utils.compat` fails to capture latest `plum` api changes


yzjin:
On line [78 and 79](https://github.com/SSAGESLabs/PySAGES/blob/3bd3f4c07d82b7a716085b11976094876a7fabfb/pysages/utils/compat.py#L78) of`pysages.utils.compat`, the function calls two functions in `plum`: `plum.get_origin()` and `plum.get_args()`. Both functions have been removed from `plum` in a recent [breaking change](https://github.com/beartype/plum/pull/94/commits/66072ddcd9825ab235e03d8ee2b3bb514e644335), and lead to `AttributeError`. These functions should probably be replaced by the native `typing.get_origin()` and `typing.get_args()`.

pabloferz:
Solved by #280.

PR #280, opened 2023-09-06
Fix compatibility issue with `plum >= 2.2.1`

yzjin:
This pr aims to fix the compatibility issue caused by `pysages/utils/compat.py`. During recent release 2.2.1, `plum` has dropped its support to Python 3.7, and removed functions and classes that were used as a workaround. As a result, I made edits to the scripts such that it will resort to Python's built-in `typing` module for `get_args()`, `get_origin()` and `Union`. These functions are introduced in Python 3.8, which is an install requirement of the latest `plum` module.

### DIFF
--- a/pysages/utils/compat.py
+++ b/pysages/utils/compat.py
@@ -1,8 +1,6 @@
 # SPDX-License-Identifier: MIT
 # See LICENSE.md and CONTRIBUTORS.md at https://github.com/SSAGESLabs/PySAGES
 
-import typing
-
 from importlib import import_module
 
 from jax.scipy import linalg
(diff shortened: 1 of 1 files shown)


pysages/utils/compat.py
```diff
@@ -66,7 +66,10 @@ def has_method(fn, T, index):
 
 else:
     _bt = import_module("beartype.door")
-    _pm = import_module("plum")
+    if _plum_version_tuple < (2, 2, 1):
+        _typing = import_module("plum")
+    else:
+        _typing = import_module("typing")
```
pabloferz, line 72:
```suggestion
    _typing = import_module("plum" if _plum_version_tuple < (2, 2, 1) else "typing")
```

pabloferz:
Fixes #278.

pabloferz (approved):

PR #292, opened 2024-01-02
Count how many times each method is called

pabloferz:
This makes sure that each sampling method has a `ncalls` field that counts how many times it's `update` method has been called.

This should make easier implementing tooling for handling workflows with restarts as we can better find out for how long has a method run.


gustavor101 (approved):
Changes work  as intended, a wall limit could be included in run method for max number of ncalls.

PR #309, opened 2024-03-11
Enable continuous GeM computation.

maggiezimon:
The continuous calculation of GeM is enabled by adding a new score as defined in Eq. A1-A2 in ["Local-order metric for condensed-phase environments"](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.97.064105).

### DIFF
--- a/pysages/colvars/patterns.py
+++ b/pysages/colvars/patterns.py
@@ -67,7 +67,7 @@ class Pattern:
         self.number_of_added_sites = number_of_added_sites
         if self.number_of_added_sites > 0:
             if width_of_switch_func is None:
-                self.width_of_switch_func = self.standard_deviation/2
+                self.width_of_switch_func = self.standard_deviation / 2
             else:
                 self.width_of_switch_func = width_of_switch_func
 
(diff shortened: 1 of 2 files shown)


pysages/colvars/patterns.py
```diff
@@ -42,6 +44,9 @@ def __init__(
         centre_j_id,
         standard_deviation,
         mesh_size,
+        number_of_added_sites=0,
```
maggiezimon, line 47: If the continuous version of the LoM is to be used, the additional atoms should already be added at the end of the `reference` (list of coordinates of the reference structure). With `number_of_added_atoms` we specify how many were included.

pabloferz:
@maggiezimon have you tried to take the gradient of GeM? E.g.
```python
cv = GeM(...)
jx.grad(cv.function)(positions)
```

maggiezimon:
I didn’t run the jax.grad. Sorry!
 
I just assumed it should work as the new score function, if number_of_added_sites > 0, returns jax.lax.erfc. Maybe I need to change how we output with the calculate_lom function.

maggiezimon:
The grad works if the positions are passed directly, instead of being read from the neighbor list.

Issue #305, opened 2024-03-02
RMSD calculation may lead to NaN values


yzjin:
The RMSD collective variable [implementation](https://github.com/SSAGESLabs/PySAGES/blob/c0fe37ee2172502963b0a2c93047a86ad5d76247/pysages/colvars/orientation.py#L366) is problematic. It leads to numerical difference in the test case used [here](https://github.com/charnley/rmsd/blob/master/tests/test_kabsch.py) and NaN values even when the weights are removed [here](https://github.com/charnley/rmsd/blob/master/tests/test_kabsch_weighted.py). The root cause of such behavior is unknown yet and require further investigation.

pabloferz:
The implementation now uses by default Kabsch algorithm as implemented in #321. Is this still happening or can we close the issue?

yzjin:
I see! Closing this issue.

pabloferz:
It's still may be worth making sure that it works against those test cases. Could you double check @yzjin?

PR #321, opened 2024-05-24
Implementation of eRMSD (AA), eRMSD (CG) and  RMSD (Kabsch) [all unweighted]

yihengwuKP:
Hi,  I've implemented [eRMSD](https://doi.org/10.1093/nar/gku972) in AA or CG mappings. I've also incorporated Yezhi's implementation of Kabsch RMSD into the code @yzjin . I've implemented tests for all of them. 

The eRMSD requires knowing the C2/C4/C6 of the six-membered ring in the base. The implementation of eRMSDCG is for the situation when we only know the other three sites in the base, and we need to first infer the position of C2/C4/C6 before we calculate the eRMSD. 

The testing of eRMSD (AA) was done by comparing with the calculation exactly done by [barnaba](https://github.com/srnas/barnaba). The testing of eRMSDCG was done by comparing with the eRMSD so it's not exact but very close. The testing for RMSD is done by precalculated examples as in [test kabsch](https://github.com/charnley/rmsd/blob/master/tests/test_kabsch.py) as mentioned in https://github.com/SSAGESLabs/PySAGES/issues/305

Please let me know if there are any suggestions/comments!


pysages/colvars/orientation.py
```diff
@@ -0,0 +1,567 @@
...
+        Weights of the selected atom positions
+    """
+    P = fitted_positions(r1, w_0)
+    U = kabsch(P, Q)
+    optimal_P = np.dot(P, U)
+    error = np.sqrt(np.sum(np.square(optimal_P - Q)) / P.shape[0])
+    return error
+
+
+class RMSD_Kabsch(CollectiveVariable):
```
pabloferz, line 96: I would prefer to have just one RMSD class and have a class argument for indicating the algorithm to estimate the optimal rotation matrix. Something along these lines:
```python
class RMSD(CollectiveVariable):
    def __init__(self, indices, references, weights=None, optimal_rotation: Callable = kabsch):
        if weights is not None and len(indices) != len(weights):
            raise RuntimeError("Indices and weights must be of the same length")
        super().__init__(indices)
        self.references = np.asarray(references)
        if weights is None:
            self.weights = np.ones(len(indices)) / len(indices)
        else:
            self.weights = np.asarray(weights)
        self.Q = fitted_positions(self.references, self.weights)
        self.optimal_rotation = optimal_rotation

    @property
    def function(self):
        return lambda r: rmsd(r, self.Q, self.weights, self.optimal_rotation)


def rmsd(r1: np.ndarray, Q: np.ndarray, w_0: np.ndarray, optimal_rotation: Callable):
    """
    Calculate the RMSD respect to a reference using rotation matrix.
    Parameters
    ----------
    r1: np.ndarray
        Atomic positions.
    Q: np.ndarray
        Cartesian coordinates of the reference position of the atoms.
    w_0: np.ndarray
        Weights of the selected atom positions
    """
    P = fitted_positions(r1, w_0)
    U = optimal_rotation(P, Q)
    optimal_P = np.dot(P, U)
    error = np.sqrt(np.sum(np.square(optimal_P - Q)) / P.shape[0])
    return error
```

pabloferz (changes requested):

pysages/colvars/orientation.py
yihengwuKP: cool! I've just commit the changes for this!

yihengwuKP:
not sure why the Docker build test is failing...

pabloferz:
Can you rebase into the latest main?

yihengwuKP:
@pabloferz can we merge this now?

pabloferz:
Yes, sorry about the delay. Thank you for adding these!

yihengwuKP:
no problem, glad to help!

Issue #323, opened 2024-06-01
Plum dispatch>2.3.0 may cause compatibility issue with PySAGES


yzjin:
A [commit](https://github.com/beartype/plum/commit/8b43135ee0a93a8a3ffb08e75b14725b7e413562) in the plum dispatch 2.3.0 contains a change that is not well documented. The `Dispatcher.methods` [attribute](https://github.com/beartype/plum/blob/8b43135ee0a93a8a3ffb08e75b14725b7e413562/plum/resolver.py#L103) is no longer a list of `signature`, but become a list of `Method`, which pushed the signature to one level below and cause compatibility issues with pysages. The `has_method()` function [here](https://github.com/yzjin/PySAGES/blob/da514637b90165c27de8e0e9d1bd255969ebd679/pysages/utils/compat.py#L112) will no longer be able to fetch the signature of the method correctly. I will create a pull request later to address this issue.

pabloferz:
Closed by #324

PR #324, opened 2024-06-01
Fix plum compatibility issue caused by 2.3.0 release

yzjin:
This pull request aims to fix the issue #323. A new signature fetching mechanism will be used whenever the `plum` version is greater than `2.3.0`.

### DIFF
--- a/pysages/utils/compat.py
+++ b/pysages/utils/compat.py
@@ -103,16 +103,20 @@ else:
     _typing = import_module("plum" if _plum_version_tuple < (2, 2, 1) else "typing")
     _util = _typing.type if _plum_version_tuple < (2, 2, 1) else _typing
 
+    if _plum_version_tuple < (2, 3, 0):
+        def _signature_types(sig):
+            return sig.types
+    else:
+        def _signature_types(sig):
+            return sig.signature.types
+
     def dispatch_table(dispatch):
         return dispatch.functions
 
     def has_method(fn, T, index):
         types_at_index = set()
         for sig in fn.methods:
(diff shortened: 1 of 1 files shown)


pabloferz:
Can you rebase these changes into the latest main?

pysages/utils/compat.py
```diff
@@ -94,7 +94,10 @@ def dispatch_table(dispatch):
     def has_method(fn, T, index):
         types_at_index = set()
         for sig in fn.methods:
-            typ = sig.types[index]
+            if _plum_version_tuple < (2, 3, 0):
+                typ = sig.types[index]
+            else:
+                typ = sig.signature.types[index]
```
pabloferz, line 100: It's more robust to do the version checking outside the function we make use of. So if you do something like this
```suggestion
            typ = _signature_types(sig)[index]
```
You can add between lines 89 and 91, the following:
```python
    if _plum_version_tuple < (2, 3, 0):

        def _signature_types(sig):
            return sig.types

    else:
        def _signature_types(sig):
            return sig.signature.types

```

pabloferz (changes requested):

pysages/utils/compat.py
yzjin: Changes incorporated into the latest version, let me know if they looks good to you.

pabloferz:
I'm a bit puzzled about the failing test I can see how this PR should impact that at all. I'll try to dig a bit into it.

pabloferz (approved):

PR #325, opened 2024-06-21
[LAMMPS] Add conversion factor depending on the unit system

pabloferz:


pabloferz:
Still WIP, but it might be interesting to look at @ramongonzalezperez, @loganhennes

PR #327, opened 2024-07-09
Implementations of 2 new cvs: native contact fraction Q and angles between two vectors

yihengwuKP:
Hi I've implemented two new cvs: 1. angle between two vectors 2. native contact fraction Q. Please let me know if anything needs to be modified! I've added another python script under colvars `contacts.py` for Q, but the vector angles are added in `angles.py`. I've implemented testing for Q and its derivatives in `example/openmm/cv_Q`.


pysages/colvars/contacts.py
```diff
@@ -0,0 +1,123 @@
...
+    index_mapping = np.zeros(old_indices.max() + 1, dtype=new_indices.dtype)
+    index_mapping = index_mapping.at[old_indices].set(new_indices)
+
+    contacts = np.array(contacts)
+    remapped_contacts = index_mapping[contacts]
+
+    return remapped_contacts
+
+
+@multicomponent
```
pabloferz, line 28: Is `@multicomponent` really needed here? AFAICT, this returns a scalar?

pysages/colvars/angles.py
```diff
@@ -128,6 +128,56 @@ def dihedral_angle(p1, p2, p3, p4):
...
+    def function(self):
+        """
+        Returns
+        --------
+        Functions that calculates the angle value from a simulation snapshot.
+        """
+        return vector_angle
+
+
+def vector_angle(p1, p2, p3, p4):
```
pabloferz, line 147: I think we should be able to reduce some code duplication by having our `angle` function taking four arguments and then change the `Angle` class call a lambda that calls `angle` with one of the arguments repeated.

pysages/colvars/contacts.py
```diff
@@ -0,0 +1,123 @@
...
+    # create the mapping from old indices to new indices
+    old_indices = np.array(sorted_unique_indices)
+    new_indices = np.arange(len(sorted_unique_indices))
+    index_mapping = np.zeros(old_indices.max() + 1, dtype=new_indices.dtype)
+    index_mapping = index_mapping.at[old_indices].set(new_indices)
+
+    contacts = np.array(contacts)
+    remapped_contacts = index_mapping[contacts]
+
+    return remapped_contacts
```
pabloferz, line 25: I need to check, but this looks like something that could be accomplished, by using `argsort`

pysages/colvars/contacts.py
yihengwuKP: good catch, I think perhaps I copied some other code with the multicomponent. I'll fix it!

pysages/colvars/angles.py
yihengwuKP: I see, that sounds cool

PR #354, opened 2025-06-25
Fixes and optimizations to jax-md backend

em819:
I have performed following fixes concerning the jax-md backend of PySAGES:

1. Changed the point in _step() at which the snapshot is updated. This addresses the corner case of performing a single MD step. In the previous implementation after a single MD step, the snapshot would still contain the original state (i.e the 0th state). Thus if one would repeatedly restart with single-step-runs, the trajectory of snapshots would not change. This is now fixed.
2. Extended the contents of the Snapshot class to also optionally contain thermostat information, which by jax-md is saved in state.chain.
3. Rewritten run-method in jax-md backend to optionally use jax.lax.fori_loop() if jit_compile=True. In addition, now the step and run function is defined/instantiated only once even if pysages.run is called multiple times in succession. If pysages.run is called repeatedly these functions are now reused and thus don't need to be jit-recompiled. Particularly for smaller systems with less than 10000 atoms, I have observed a speed-up of up to factor 10 compared to the original version.

### DIFF
--- a/pysages/backends/jax-md.py
+++ b/pysages/backends/jax-md.py
@@ -123,7 +123,11 @@ def build_runner(context, sampler, jit_compile=True):
 
         def _run_body(i, input_states_and_snapshots):
             context_state, snapshot, sampler_state = input_states_and_snapshots
-            return tuple(step(context_state, snapshot, sampler_state))
+            context_state, snapshot, sampler_state = step(context_state, snapshot, sampler_state)
+
+            if sampler.callback:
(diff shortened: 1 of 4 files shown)


em819:
Hey @pabloferz,
Thanks for having a look, much appreciated.

p-zubieta:
I was thinking on doing it on our end, cherry-picking your commits. We did some internal changes to allow for the inclusion of additional data into Snapshots a bit easier and want to move this on top of those.

em819:
Sure thing, sounds good.

